PR #6, opened 2019-10-18
Add explanation about av-sync behavior

kuddai:
Addresses this [issue](https://github.com/henbos/webrtc-extensions/issues/5).

### DIFF
--- a/index.html
+++ b/index.html
@@ -110,6 +110,19 @@
             influence the User Agent to buffer these packets rather than playing
             them out.
           </p>
+          <p class="note">
+            If the track is paired with other tracks through
+            <code><a>RTCRtpReceiver</a></code>
+            <a href="https://w3c.github.io/webrtc-pc/#dfn-associatedremotemediastreams">
+            [[\AssociatedRemoteMediaStreams]]</a> internal slot, then it will
+            be synchronized with other tracks (for e.g. audio video
+            synchronization). This means that even if one of the paired tracks
+            is delayed through [[\PlayoutDelayHint]] then the User Agent
+            synchronization mechanism will automatically delay all others
+            paired tracks. If multiple such paired tracks are delayed through
+            <a>[[\PlayoutDelayHint]]</a> by different amounts then the largest
+            of those hints will take precedence in synchronization mechanism.
+          </p>
           <p>On getting, this attribute MUST return the value of the
           <a>[[\PlayoutDelayHint]]</a> internal slot.</p>
           <p>On setting, the User Agent MUST run the following steps:</p>


index.html
```diff
@@ -110,6 +110,19 @@ <h2>Attributes</h2>
...
+            <code><a>RTCRtpReceiver</a></code>
+            <a href="https://w3c.github.io/webrtc-pc/#dfn-associatedremotemediastreams">
+            [[\AssociatedRemoteMediaStreams]]</a> internal slot, then it will
+            be synchronized with other tracks (for e.g. audio video
+            synchronization). This means that even if one of the paired tracks
+            is delayed through [[\PlayoutDelayHint]] then the User Agent
+            synchronization mechanism will automatically delay all others
+            paired tracks. If multiple such paired tracks are delayed through
+            <a>[[\PlayoutDelayHint]]</a> by different amounts then the largest
+            of those hints will take precedence in synchronization mechanism.
```
henbos: Just confirming: We pick the highest value, not the lowest? (Either is fine by me)

kuddai: I have checked it on appr.tc demo app, it picks up the highest delay. I haven't found any spec which would describe mechanism of **how** streams must be synchronized, so if implementation of a/v sync changes then this paragraph won't be true anymore.

henbos: The way I thought of it was that it would use the highest hint value and pass that down to both receivers' jitter buffers. But if A/V sync makes it work implicitly that's fine too?

kuddai: At the level a/v sync is implemented, both a/v sync and playoutDelayHint is relaying on similar jitter Buffer API.

playoutDelayHint affects jitter buffer through [webrtc::Delayable](https://cs.chromium.org/chromium/src/third_party/webrtc/media/base/delayable.h?l=29&rcl=b9f69028a0dab597c3d137d2bc549fcf475bc66b
)

a/v [sync](https://cs.chromium.org/chromium/src/third_party/webrtc/video/rtp_streams_synchronizer.cc?l=117&rcl=b9f69028a0dab597c3d137d2bc549fcf475bc66b) changes it through [webrtc::Syncable](https://cs.chromium.org/chromium/src/third_party/webrtc/call/syncable.h?l=39&rcl=b9f69028a0dab597c3d137d2bc549fcf475bc66b)


Then among the two, the maximum value is used.

Here for e.g. the maximum is taken for [video](https://cs.chromium.org/chromium/src/third_party/webrtc/video/video_receive_stream.cc?l=706&rcl=b9f69028a0dab597c3d137d2bc549fcf475bc66b) and here is for [audio](https://cs.chromium.org/chromium/src/third_party/webrtc/modules/audio_coding/neteq/delay_manager.cc?l=508&rcl=b9f69028a0dab597c3d137d2bc549fcf475bc66b).

PR #7, opened 2019-10-22
Throw a TypeError instead of InvalidAccessError

foolip:
InvalidAccessError is deprecated:
https://heycam.github.io/webidl/#invalidaccesserror


foolip:
This is like https://github.com/henbos/webrtc-timing/pull/5, don't know which spec is for realz :)

henbos:
This is the fo' realsies spec!

Issue #12, opened 2019-10-29
Testability of playoutDelayHint


henbos:
While the hint may be ignored in various circumstances, there should be enough specificity here to write some basic web platform tests.

Such as measuring jitterBufferDelay / jitterBufferEmittedCount and ensuring the value is &gt;= the hint is achieved, etc. Let's write WPTs and, if anything needs to be clarified or reworded in the spec to ensure the test is interoperable with the spec definition, we do that too.

@kuddai

henbos:
Related to #10, but we can wait that within reasonable testing timeouts we should eventually see that the hint is achieved under ideal circumstances.

fippo:
single-page tests are actually ideal circumstances - we can only test those however (at least in WPT).
Not sure if you can use the WebRTCFakeNetwork field trials in chrome tests.

Made http://fippo.github.io/webrtc//src/content/peerconnection/playoutDelayHint/index.html for playing with this (one needs to manually set in the console, e.g. `pc2.getReceivers()[1].jitterBufferDelayHint = 2.2. Looks like the stats are not on the right object yet because you axed track stats? Otherwise this should be somewhat testable.

henbos:
> Looks like the stats are not on the right object yet because you axed track stats?

A bunch of stuff recently moved in the spec from "track" to "outbound-rtp" (to support simulcast) and "inbound-rtp" (for consistency). "track" should be removed / return duplicate information for backwards-compatiblity reasons, but we've not updated out in/outbound-rtps yet.

henbos:
Cool, here's the https version of the link or else gUM fails ;) https://fippo.github.io/webrtc//src/content/peerconnection/playoutDelayHint/index.html

henbos:
In PR #15 we clarified that this can be measured with jitterBufferDelay/jitterBufferEmittedCount, so we're almost ready to close this issue.

But @jan-ivar [asked](https://github.com/henbos/webrtc-extensions/pull/15#discussion_r348249295) "Can we put any limits on how long implemenations are allowed to take here, to make this testable?" and @kuddai replied (see link) that it is hard to say and depends on network conditions, speech type, etc. Chrome's implementation looks at the wave lengths and adjusts samples in a way as to minimize the perceived effect.

I think we just want to get across that a) it may take several seconds for this (so that an implementation does not cut it off and cause stutter), and b) it should not take several minutes. Chrome may take around 20 s, but this area is still very much being experimented on, so we're reluctant to say something very exact.

[...]

Issue #10, opened 2019-10-29
Clarify what to do when playoutDelayHint changes its value


henbos:
Per discussions at #8, there could be a compat issue e.g. if one implementation just drops all the audio/video when decreasing the buffer size but another accelerates or decelerates.

We should add guidance to inform implementers what to do when the hint changes.

I believe the intent is to perform acceleration or deceleration? @kuddai

henbos:
> For instance, if the goal of changing the value from 10 to 3 is not to drop 7 seconds of frames instantly, we should say so. Or if that is the goal, we should say that instead. Once we expose a web control surface, I think it helps web compat to specify how it works.

henbos:
With PR #15 and follow-up issues #11 and #12 I think we can close this issue.

Issue #11, opened 2019-10-29
Specify upper limit to playoutDelayHint


henbos:
Per discussions at #8, it would be good for specificity, interoperability and testability if we had an explicit upper limit to the hint. We don't want this API to be a way to buffer data for 10 minutes for example.

henbos:
@kuddai [said](https://github.com/henbos/webrtc-extensions/issues/8#issuecomment-546965126) that current implementation in Chrome has "4 seconds for audio buffer and 10 seconds for video buffer"

How about we make the upper limit 4 seconds for both audio or video?

kuddai:
Starting lower is the safer because if we ever want to increase the maximum then it won't break existing implementations that relying on that behavior.

What error type would be appropriate if user sets value beyond maximum allowed one?

@jan-ivar what do you think about 4 second as maximum value?

henbos:
In webrtc-pc we use RangeError and InvalidModificationError. I guess if it's a hardcoded limit, RangeError makes sense, unless there are historical reasons for preferring one exception over another.

PR #15, opened 2019-11-14
Add information how to achieve preferred buffering level.

### LINKED ISSUE #9: Clarify that playoutDelayHint is expressed in absolute value, not relative
Per discussions at #8, it may not be clear that you set the hint to the absolute delay that you want, not additional delay on top of implementation-specific buffering. We should add a comment to make this clear.

@kuddai can I assign this to you?

kuddai:
Fixes #9, #10, #12

### DIFF
--- a/index.html
+++ b/index.html
@@ -172,7 +172,7 @@
             of those hints will take precedence in synchronization mechanism.
           </p>
           <p class="note">
-            Receiver's current delay can be measured as the delta
+            The receiver's average delay can be measured as the delta
             <code><a href="https://w3c.github.io/webrtc-stats/#dom-rtcinboundrtpstreamstats-jitterbufferdelay">
             jitterBufferDelay</a></code> divided by the delta
             <code><a href="https://w3c.github.io/webrtc-stats/#dom-rtcinboundrtpstreamstats-jitterbufferemittedcount">
--- a/index.html
+++ b/index.html
@@ -140,18 +140,20 @@
         <dt><dfn><code>playoutDelayHint</code></dfn> of type <span
           class="idlAttrType">double</span>, nullable</dt>
         <dd>
-          <p>This attribute allows the application to give a hint to the User
-          Agent about what total delay is acceptable before the playout. The
+          <p>This attribute allows the application to specify a target duration
+          of time between network reception of media and playout. The
           User Agent SHOULD NOT playout audio or video that is received unless
-          this total amount of time has passed in seconds, allowing the User
+          this amount of time has passed in seconds, allowing the User
           Agent to perform more or less buffering than it might otherwise do.
           This allows to influence the tradeoffs between having a higher delay
           and the risk that buffers such as the jitter buffer will run out of
-          audio or video frames to play due to network jitter. This API only
-          serves to provide a hint, and the User Agent may ignore it. The hint
-          SHOULD NOT be followed if it significantly impacts audio or video
-          quality (e.g. bad network), or if the value implies allocating larger
-          buffers than the User Agent is willing to provide.</p>
+          audio or video frames to play due to network jitter. 
+          If the given delay value is larger than the User Agent is able or
+          willing to provide, the User Agent MUST use the maximum value
+          available to match the requested delay as close as possible. If at
+          any point in time delay is smaller than the User Agent determines
+          is needed in order provide sufficient audio or video quality, the User
+          Agent MUST use the smallest value possible.</p>
           <p class="note">
             The playout delay hint applies even if DTX is used. For example, if
             DTX is used and packets start flowing after silence, the hint can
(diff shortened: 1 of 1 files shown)


kuddai:
Addresses issue ["Clarify what to do when playoutDelayHint changes its value #10"](https://github.com/henbos/webrtc-extensions/issues/10).

kuddai:
I think the side note about how fast the preferred buffering level is achieved can be improved or removed completely. It is indeed an implementation detail as if your networks conditions are poor and packets are lost or delayed, then time to achieve preferred buffering level will be very different from the time in perfect network conditions. It also depends on the codec types, as for e.g. different audio codecs may provide different capabilities for seamless audio acceleration/deceleration. 

In Chrome by the time of writing the maximum audio acceleration/deacceleration is ~15% from original audio duration. The exact mechanism used in Chrome is the same one which is used for A/V sync, congestion control and [playout delay rtp header extension](https://webrtc.org/experiments/rtp-hdrext/playout-delay/), but of course it doesn't mean that other User agents should do the same internally as it is implementation detail.

In our origin trial experiments when we were dynamically changing hint depending on users behavior we were also changing its gradually so the change in pacing playout remained unnoticeable to the users.

index.html
```diff
@@ -93,23 +93,30 @@ <h2>Attributes</h2>
...
-          buffering than it might otherwise do. This allows to influence the
-          tradeoffs between having a higher delay and the risk that buffers such
-          as the jitter buffer will run out of audio or video frames to play due
-          to network jitter. This API only serves to provide a hint, and the
-          User Agent may ignore it. The hint SHOULD NOT be followed if it
-          significantly impacts audio or video quality (e.g. bad network), or if
-          the value implies allocating larger buffers than the User Agent is
-          willing to provide.</p>
+          Agent about what total delay is acceptable before the playout. The
+          User Agent SHOULD NOT playout audio or video that is received unless
```
henbos: Separate from this PR, but if we can specify that up to X seconds of delay must be supported then we might be able to make this language a bit stronger (change a SHOULD to a MUST), and say that anything above X just throws.

index.html
```diff
@@ -93,23 +93,30 @@ <h2>Attributes</h2>
...
+          quality (e.g. bad network), or if the value implies allocating larger
+          buffers than the User Agent is willing to provide.</p>
           <p class="note">
             The playout delay hint applies even if DTX is used. For example, if
             DTX is used and packets start flowing after silence, the hint can
             influence the User Agent to buffer these packets rather than playing
             them out.
           </p>
+          <p class="note">
+            Receiver's current delay can be measured as the delta
```
henbos: nit: The receiver's average delay for an interval can be measured as the delta...

index.html
```diff
@@ -149,6 +156,21 @@ <h2>Attributes</h2>
                     reflected in the <a>[[\PlayoutDelayHint]]</a> internal slot.
                   </p>
                 </li>
+                <li>
+                  <p>Modifying the delay of the underlying system SHOULD affect
+                  the internal audio or video buffering gradually in order not
+                  to hurt user experience. Audio samples or video frames SHOULD
+                  be accelerated or decelerated before playout, similarly to how
+                  it is done for audio/video synchronization or in response to
+                  congestion control.</p>
```
henbos: "similarly to how it is done" is probably a step in the right direction. This shows that there is room for improvement in specificity when it comes to acceleration and deceleration, but I don't think this PR is the place to define general principles of acceleration and deceleration. Ideally there would be an acceleration or deceleration spec already that we could just reference, since we are reusing existing mechanisms rather than invent our own. For now I like it, but I recognize the need for something like a separate acceleration/deceleration spec.

I think this PR answers the broader question about "what to do?", namely that you want a smooth transition.

kuddai: Agree, currently we have a bit fragmented picture. Acceleration or deceleration don't exist as separate concepts but similarly to jitter buffer their existence is heavily hinted from WebRTC stats. Lip sync is also a good doc as synchronization implementation originates through those SDP named groups.

index.html
```diff
@@ -141,17 +141,17 @@ <h2>Attributes</h2>
...
+          User Agent SHOULD NOT playout audio or video that is received unless
+          this total amount of time has passed in seconds, allowing the User
+          Agent to perform more or less buffering than it might otherwise do.
+          This allows to influence the tradeoffs between having a higher delay
+          and the risk that buffers such as the jitter buffer will run out of
+          audio or video frames to play due to network jitter. This API only
+          serves to provide a hint, and the User Agent may ignore it. The hint
+          SHOULD NOT be followed if it significantly impacts audio or video
+          quality (e.g. bad network), or if the value implies allocating larger
+          buffers than the User Agent is willing to provide.</p>
```
jan-ivar: I wonder if we can more specifically say that user agents must define a _floor value_ and a _ceiling value_ to represent these properties.

That way we can put normative requirements on that user agents MUST respect values in between those self-defined limits. That seems like the right trade-off for web compat to me.

Update: I see the prose steps below sort of have these concepts already, but maybe we can formalize it as definitions?

index.html
```diff
@@ -210,6 +217,23 @@ <h2>Attributes</h2>
...
                 </li>
+                <li>
+                  <p>Modifying the delay of the underlying system SHOULD affect
+                  the internal audio or video buffering gradually in order not
+                  to hurt user experience. Audio samples or video frames SHOULD
+                  be accelerated or decelerated before playout, similarly to how
+                  it is done for
+                  <a href="https://tools.ietf.org/html/rfc5888#section-7">
+                  audio/video synchronization</a> or in response to congestion
+                  control.</p>
```
jan-ivar: Can we put any limits on how long implemenations are allowed to take here, to make this testable?

kuddai: Overall time needed to achieve depends on the network conditions. Network delays and packet's drops can heavily influence overall time between receiving and playout. 

Theoretically, we can specify  how fast it is allowed to stretch (deaccelerate) and shorten (accelerate) audio and video as these parameters are essentially putting margins on "don't hurt User Experience".  This would allow to specify how fast we can achieve target in the case of "perfect network conditions" which in practice is the localhost connection.

But at least in libwebrtc implementation those parameters are actively adjusted through experimentation and most importantly in audio case they are tied up with the type of the audio. For example you can accelerate almost instantly if your audio frame contains lots of silence/minor background noise. If it contains speech then speech related traits play major role such as peeks, pace, timber, etc. Essentially, shape of the waveform. Thus, even for localhost testing you would get different results depending on audio being used.

I want to emphasize that this testing issue would be present independently from the API path chosen.

kuddai: We can try to provide maximum time beyond which should be considered a bug in "perfect network" conditions. It should give some reference for the vendors. But I would advise against it. I can measure the Chrome speed with some constant speech feed and take x2.0 of that time to estimate the maximum allowed margin.

henbos: Let's follow up on this in #11.

Issue #20, opened 2019-11-25
The naming of playoutDelayHint


henbos:
@jan-ivar suggested renaming it to playoutDelay.

We are OK with this, but let's make sure we want this, we don't want to go back and forth :)

henbos:
With playoutDelayHint becoming more well-defined and testable, playoutDelay has my vote.

padenot:
`playoutDelay` sounds like the total end-to-end latency. This clearly refers to the jitter buffer duration target to do a trade-off between latency and resilience.

jan-ivar:
It's still called _playoutDelayHint_ in Canary FWIW (typing into web console `new RTCPeerConnection().addTransceiver("audio").receiver.playoutDelayHint` produces `null`).

PR #22, opened 2019-11-25
Update name, add limits on maximum value and speed.

### LINKED ISSUE #11: Specify upper limit to playoutDelayHint

kuddai:
Fixes #11, #12, #20 .

### DIFF
--- a/index.html
+++ b/index.html
@@ -196,7 +196,7 @@
             </li>
             <li>
               <p>If <var>delay</var> is negative or larger than 4 seconds then,
-              <a>throw</a> an <code>TypeError</code> and abort these steps.</p>
+              <a>throw</a> an <code>RangeError</code> and abort these steps.</p>
             </li>
             <li>
               <p>Set the value of <var>receiver</var>'s <a>[[\PlayoutDelay]]</a>
--- a/index.html
+++ b/index.html
@@ -131,17 +131,17 @@
     </p>
     <p>
       Let <code>RTCRtpReceiver</code> objects have a
-      <dfn>[[\PlayoutDelayHint]]</dfn> internal slot initially
+      <dfn>[[\PlayoutDelay]]</dfn> internal slot initially
       initialized to <code>null</code>.
     </p>
     <pre class="idl">partial interface RTCRtpReceiver {
-  attribute double? playoutDelayHint;
+  attribute double? playoutDelay;
   };</pre>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -237,6 +236,11 @@ <h2>Attributes</h2>
                     and <code><a href="https://w3c.github.io/webrtc-stats/#dom-rtcinboundrtpstreamstats-removedsamplesforacceleration">removedSamplesForAcceleration</a></code>.
                     For video, this may result in the same frame being rendered
                     multiple times or frames may be dropped.
+
+                    With constant audio/video feed received under the perfect
+                    network conditions, for e.g. loopback connection over
+                    localhost, it SHOULD take less than 30 seconds on average to
+                    reach 1 second buffering level from empty buffer.
```
henbos: Do we really need the "perfect network conditions" part? In a real-world use case you wouldn't have perfect conditions, so if the spec only mentions the perfect case then it's not as clear how this is to be used by real applications.

Can this say "The acceleration or deceleration rate may vary depending on network conditions or the type of audio received (e.g. speech or background noise). It MAY take several seconds to achieve 1 second of buffering but SHOULD not take more than 30 seconds assuming packets are being received. The speed MAY be different for audio and video."

Also can you move this to the previous paragraph (outside the "note")?

index.html
```diff
@@ -195,13 +195,12 @@ <h2>Attributes</h2>
               <p>Let <var>delay</var> be the argument to the setter.</p>
             </li>
             <li>
-              <p>If <var>delay</var> is negative, <a>throw</a> an
-              <code>TypeError</code> and abort these steps.</p>
+              <p>If <var>delay</var> is negative or larger than 4 seconds then,
```
henbos: @foolip You changed InvalidAccessError to TypeError. However, doesn't RangeError make more sense?

index.html
kuddai: Done.

kuddai:
@jan-ivar can you please take a look?

index.html
foolip: Yeah, that'd work too.

kuddai: Change it to RangeError.

PR #25, opened 2020-01-28
Add RTP header negotiation control.

### LINKED ISSUE #24: Add API for controlling RTP header extension negotiation and usage
Discussion document:

https://docs.google.com/document/d/1y1hTsMeav5ijPvoqu1R6U4YC564i1QzgkMeIqWhgiis/edit#

Approach presented to WG in January 2020, and approved for writing a PR in this repo.

alvestrand:
Fixes #24

Note: Initial PR has SDP control surface only.

### DIFF
--- a/index.html
+++ b/index.html
@@ -201,11 +201,12 @@ sequence&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer);
     </pre>
     <p>
       Let [[\HeaderExtensionsToOffer]] be an internal slot of the RTCRtpTransceiver,
-      initialized to the platform-specific initial set of RTP header extensions that
-      is offered on an initial CreateOffer. This list MUST include all header extensions
-      that are marked Mandatory to Implement in the underlying protocol. The direction
+      initialized to the platform-specific list of implemented RTP header extensions.
+      The direction
       attribute for all extensions that are mandatory to use MUST be initialized to
-      an appropriate value other than "stopped".
+      an appropriate value other than "stopped". The direction attribute for extensions
+      that will not be offered by default in an initial offer MUST be initialized to
+      "stopped".
     </p>
     <p class="note">The list of header extensions that MUST/SHOULD be supported is listed in
       [[RTCWEB-RTP]], section 5.2. The "mid" extension is mandatory to use when
@@ -263,24 +264,25 @@ sequence&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer);
             <li>
               For each extension <var>ext</var> in <var>extlist</var>, do the following:
               <ol>
-                <li>If <var>ext</var>'s "uri" value is not a supported extension, throw
-                  a NotSupported error.
+                <li>If <var>ext</var>'s "uri" value is not present in
+                  [[\HeaderExtensionsToOffer]], throw a NotSupported error.
                 </li>
-                <li>If <var>ext</var>'s "direction" value is not "sendrecv", and "uri"
+                <li>
+                  Let <var>entry</var> be the entry with the same "uri" value in
+                  [[\HeaderExtensionsToOffer]].
+                </li>
+                <li>If <var>ext</var>'s "direction" field is not "sendrecv", and "uri"
                   indicates a mandatory-to-use attribute that is required to be both
                   sent and received, throw an IllegalArgument error.
                 </li>
-                <li>if <var>ext</var>'s "direction" value is "stopped", and "uri" indicates
+                <li>if <var>ext</var>'s "direction" field is "stopped", and "uri" indicates
                   a mandatory-to-implement extension, throw an IllegalArgument error.
                 </li>
+                <li>
+                  Set the "direction" field of <var>entry</var> to <var>ext</var>'s
+                  "direction" field.
+                </li>
               </ol>
-            </li>
-            <li>
-              For each mandatory-to-implement extension <var>mext</var>, if <var>mext</var>
-              is not present in <var>extlist</var>, throw an IllegalArgument error.
-            </li>
-            <li>
-              Set the [[\HeaderExtensionsToOffer]] to <var>extlist</var>.
             </li>
           </ol>
         </dd>
--- a/index.html
+++ b/index.html
@@ -216,43 +216,45 @@ sequence&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer);
       Let [[\HeaderExtensionsNegotiated]] be an internal slot of the RTCRtpTransceiver,
       initialized to an empty list.
     </p>
-    <p>
-      In the algorithms for generating initial offers in [[RTCWEB-JSEP]] section 5.1.1,
-      replace "for each supported RTP header extension, an "a=extmap" line, as specified in
-      [[RFC5285]], section 5" " with "For each RTP header extension "e"
-      listed in [[\HeaderExtensionsToOffer]] where direction is not "stopped", an "a=extmap"
-      line, as specified in [[RFC5285]], section 5, with direction taken from "e"'s direction
-      attribute."
-    </p>
-    <p>
-      In the algorithm for generating subsequent offers in [[RTCWEB-JSEP]] section 5.2.2, replace "The
-      RTP header extensions MUST only include those that are present in the most recent answer"
-      with "For each RTP header extension listed in [[\HeaderExtensionsToOffer]] that is
-      also present in the most recent answer, and where direction is not "stopped", generate
-      an appropriate "a=extmap" line with "direction" set according to the rules of [[RFC5285]]
-      section 6, considering the direction in [[\HeaderExtensionsToOffer]] to indicate the
-      answerer's desired usage". This preserves the property that the set of extensions
-      may not grow with subsequent offers.
-    </p>
-    <p>
-      In the algorithm for generating initial answers in [[RTCWEB-JSEP]] section 5.3.1, replace "For
-      each supported RTP header extension that is present in the offer" with "For each
-      supported RTP header extension that is present in the offer and is also present in
-      [[\HeaderExtensionsToOffer]] with a direction different from "stopped"".
-    </p>
-    <p class="note">
-      Since JSEP does not know about WebRTC internal slots, merging this change requires
-      more work on a JSEP revision.
-    </p>
-    <p>
-      In the algorithm to <a>set an RTCSessionDescription</a>, add a bullet in the sections
-      for processing answer and pranswer (bullet 3.2.8.1.7 and 3.2.8.2.6):
-      <ul>
-        <li>Set the transceiver's [[\HeaderExtensionsNegotiated]] slot to the list of
-          RTP header extensions from the answer.
-        </li>
-      </ul>
-    </p>
+    <section>
+      <h2>Modifications to existing procedures</h2>
+      <p>
+        In the <a>set an RTCSessionDescription</a> algorithm, add a step
+        right after the step that sets transceiver.[[\Sender]].[[\SendCodecs]],
+        saying "For each transciever, set [[\HeaderExtensionsNegotiated]] to
+        an empty list, and then
+        for each "a=hdrext" line, add an appropriate
+        RTCRtpHeaderExtensionCapability to the list.
+      </p>
+      <p>
+        In the algorithms for generating initial offers in [[RTCWEB-JSEP]] section 5.2.1,
+        replace "for each supported RTP header extension, an "a=extmap" line, as specified in
+        [[RFC5285]], section 5" " with "For each RTP header extension "e"
+        listed in [[\HeaderExtensionsToOffer]] where direction is not "stopped", an "a=extmap"
+        line, as specified in [[RFC5285]], section 5, with direction taken from "e"'s direction
+        attribute."
+      </p>
+      <p>
+        In the algorithm for generating subsequent offers in [[RTCWEB-JSEP]] section 5.2.2, replace "The
+        RTP header extensions MUST only include those that are present in the most recent answer"
+        with "For each RTP header extension listed in [[\HeaderExtensionsToOffer]] that is
+        also present in the most recent answer, and where direction is not "stopped", generate
+        an appropriate "a=extmap" line with "direction" set according to the rules of [[RFC5285]]
+        section 6, considering the direction in [[\HeaderExtensionsToOffer]] to indicate the
+        answerer's desired usage". This preserves the property that the set of extensions
+        may not grow with subsequent offers.
+      </p>
+      <p>
+        In the algorithm for generating initial answers in [[RTCWEB-JSEP]] section 5.3.1, replace "For
+        each supported RTP header extension that is present in the offer" with "For each
+        supported RTP header extension that is present in the offer and is also present in
+        [[\HeaderExtensionsToOffer]] with a direction different from "stopped"".
+      </p>
+      <p class="note">
+        Since JSEP does not know about WebRTC internal slots, merging this change requires
+        more work on a JSEP revision.
+      </p>
+    </section>
     <section>
       <h2>Methods</h2>
       <dl>
@@ -264,6 +266,9 @@ sequence&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer);
             <li>
               For each extension <var>ext</var> in <var>extlist</var>, do the following:
               <ol>
+                <li>
+                  If <var>ext</var>'s "uri" value is missing, throw a TypeError.
+                </li>
                 <li>If <var>ext</var>'s "uri" value is not present in
                   [[\HeaderExtensionsToOffer]], throw a NotSupported error.
                 </li>
@@ -298,26 +303,6 @@ sequence&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer);
         <dd>Returns the value of the [[\HeaderExtensionsNegotiated]] slot.</dd>
       </dl>
     </section>
-    <section>
-      <h2>Modifications to existing procedures</h2>
-      <p>
-        In the <a>final steps to create an offer</a>, add a step:
-      </p>
-      <ol>
-        <li>If the [[\HeaderExtensionsToOffer]] internal slot is set, for
-          each extension in [[\HeaderExtensionsToOffer]], add an appropriate
-          "a=hdrext" line.
-        </li>
-      </ol>
-      <p>
-        In the <a>set an RTCSessionDescription</a> algorithm, add a step
-        right after the step that sets transceiver.[[\Sender]].[[\SendCodecs]],
-        saying "For each transciever, set [[\HeaderExtensionsNegotiated]] to
-        an empty list, and then
-        for each "a=hdrext" line, add an appropriate
-        RTCRtpHeaderExtensionCapability to the list.
-      </p>
-    </section>
   </section>
   <section id="rtcrtpheaderextensionparameters-interface">
     <h3>
@@ -338,7 +323,7 @@ sequence&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer);
       <dl>
         <dt>enabled, of type boolean</dt>
         <dd>
-          <p>When returned from getParameters() on an RTPSender, "enabled" is true
+          <p>When returned from getParameters() on an RTCRTPSender, "enabled" is true
             when the RTP sender is
             configured to send this header extension when appropriate. If the
             attribute is missing, it means that it is enabled and cannot be disabled.
@@ -349,22 +334,26 @@ sequence&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer);
             to false instructs the RTP sender to never send this extension.
           </p>
           <p>
-            When calling getParameters() on an RTPReceiver, the "enabled" member will
-            always be missing, indicating that there is no control surface here.
+            When calling getParameters() on an RTCRTPReceiver, the "enabled" member will
+            always be missing.
           </p>
         </dd>
       </dl>
       <p class="note">
         The list of extensions returned from getParameters() on a sender will only include
-        the extensions that have been negotiated for sending. Changing the
-        list, apart from setting the "enabled" member, is not permitted.
+        the extensions that have been negotiated for sending.
       </p>
     </section>
     <p>
       The inclusion of a settable member of RTCRtpHeaderExtensionParameters
-      means that it is no longer a read-only member of RTCRtpParameters.
+      means that "headerExtensions" is no longer a read-only member of RTCRtpParameters.
       The "uri", "id" and "encrypted" members of
       RTCRtpHeaderExtensionParameters remain read-only.
+    </p>
+    <p>
+      Changing the
+      list, apart from setting the "enabled" member, MUST cause setParameters() to reject
+      with an InvalidModificationError.
     </p>
   </section>
   <section id="rtcrtpreceiver-interface">
--- a/index.html
+++ b/index.html
@@ -60,8 +60,25 @@
   <section id="introduction">
     <h2>Introduction</h2>
     <p>
-      TODO(henbos): Fill in.
+      This document contains proposed extensions to the [[WEBRTC]] specification,
+      where it was thought inappropriate to incorporate them directly into that specification
+      at the point in time when they were written, but they were too small to warrant
+      writing a separate document.
     </p>
+    <p>
+      If, at a later time, it is a good time to add them to the base specification, that
+      will be done. Deciding factors will include maturity of the extension, consensus on
+      adding it, and implementation experience.
+    </p>
+    <p>
+      This document contains a number of sections that have extensions for one specific
+      interface or dictionary in the base specification. When an extension only affects
+      one interface or dictionary, it will only be described there. If an extension affects
+      multiple interfaces or dictionaries, there will be a subsection in the informative
+      "Overviews" section that describes the extension as a whole, while the normative changes
+      are in the sections for the individual interfaces.
+    </p>
+    
   </section>
   <section>
     <h2>Terminology</h2>
@@ -75,6 +92,26 @@
       defined in [[WEBIDL]].</p>
     <p>The <dfn>OAuth Client</dfn> and <dfn>Authorization Server</dfn> roles
       are defined in [[RFC6749]] Section 1.1.</p>
+  </section>
+  <section id="conformance">
+  </section>
+  <section id="overviews">
+    <h2>
+      Overviews
+    </h2>
+    <section id="rtp-header-control">
+      <h3>RTP header control</h3>
+      <p>
+        The extension for RTP header extension control consists of 2 parts:
+      </p>
+      <ul>
+        <li>An extension to RTCRtpTransceiver, to set and query the RTP header extensions
+          negotiated in SDP.
+        </li>
+        <li>An extension to RTCRtpHeaderExtensionParameters, to be used with setParameters()
+          to set the RTP header extensions sent by an RTCRtpSender.</li>
+      </ul>
+    </section>
   </section>
   <section id="rtcpeerconnection-interface">
     <h3>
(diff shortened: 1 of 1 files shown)


alvestrand:
Now I think the interface is complete (but spec linkages are still not done). Comments welcome!

alvestrand:
Note: WebIDL DOES permit partial dictionary.

alvestrand:
Ping reviewers. It's Thursday tomorrow.
@jan-ivar @henbos

index.html
```diff
@@ -122,6 +159,161 @@ <h2>Methods</h2>
...
+        <dt>enabled, of type boolean</dt>
+        <dd>
+          <p>When returned from getParameters(), true when the RTP sender is
+            configured to send this header extension when appropriate. If the
+            attribute is missing, it means that it cannot be disabled.
+          </p>
+          <p>
+            When passed to setParameters(), true when the RTP sender should
+            send this header extension when appropriate; false when the RTP
+            sender should never send this extension.
```
henbos: Should this be settable? If the header extension is not negotiated and we pass "true" this needs to throw.

index.html
```diff
@@ -122,6 +159,161 @@ <h2>Methods</h2>
...
+        </dd>
+      </dl>
+    </section>
+    <p>
+      The inclusion of a settable member of RTCRtpHeaderExtensionParameters
+      means that it is no longer a read-only member of RTCRtpParameters.
+      However, the "uri", "id" and "encrypted" members of
+      RTCRtpHeaderExtensionParameters are now read-only.
+    </p>
+
```
henbos: nit: remove empty line

index.html
```diff
@@ -122,6 +159,161 @@ <h2>Methods</h2>
...
+    </h3>
+    <p>
+      The <code><a href="https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver">RTCRtpTransceiver</a></code>
+      interface is defined in [[WEBRTC]]. This document extends that interface
+      by adding an additional method and attribute in order to control negotiation of
+      RTP header extensions.
+    </p>
+    <pre class="idl">
+partial dictionary RTCRtpHeaderExtensionCapability {
+  RTCRtpTransceiverDirection direction = "sendrecv";
```
henbos: I think the implications of the direction in the context of SDP and in the context of the functionality needs to be explained.
If the direction is not included in the SDP then I would think that achieving asymmetric direction of a header extension makes more sense to do through multiple transceivers.

henbos: I.e. piggyback on the a=direction attribute of an m= section rather than inventing a RTP header extension version of the direction that needs to be added to the SDP

index.html
```diff
@@ -76,6 +93,26 @@ <h2>Terminology</h2>
...
+    </h2>
+    <section id="rtp-header-control">
+      <h3>RTP header control</h3>
+      <p>
+        The extension for RTP header extension control consists of 2 parts:
+      </p>
+      <ul>
+        <li>An extension to RTCRtpTransceiver, to set and query the RTP header extensions
+          negotiated in SDP.
+        </li>
```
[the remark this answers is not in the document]
alvestrand: HTML doesn't care, I don't care, we should be in the habit of running tidy across the spec once in a while; it cares.

index.html
```diff
@@ -122,6 +159,161 @@ <h2>Methods</h2>
...
+    <p class="note">
+      Since JSEP does not know about WebRTC internal slots, merging this change requires
+      more work on a JSEP revision.
+    </p>
+    <p>
+      In the algorithm to <a>set an SDP description</a>, add a bullet in the sections
+      for processing answer and pranswer (bullet 3.2.8.1.7 and 3.2.8.2.6):
+      <ul>
+        <li>Set the transceiver's [[\HeaderExtensionsNegotiated]] slot to the list of
+          RTP header extensins from the answer.
```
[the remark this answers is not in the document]
alvestrand: There's already a comment in the webrtc spec about which directions apply to sender vs receiver.
We can say the same thing here - the details of when to send various extensions are either in some IETF spec or left to the implementation, so what we can say here is that we enable sending (enabled=true on the sender).
The receiver has no choice about whether to receive or not, but it can choose to act only on those that have been negotiated (discard any others).

index.html
```diff
@@ -122,6 +159,161 @@ <h2>Methods</h2>
...
+      BUNDLE is in use.
+    </p>
+    <p>
+      Let [[\HeaderExtensionsNegotiated]] be an internal slot of the RTCRtpTransceiver,
+      initialized to an empty list.
+    </p>
+    <p>
+      In the algorithms for generating initial offers in JSEP section 5.1.1,
+      read "for each supported RTP header extension" as "For each RTP header extension
+      listed in [[\HeaderExtensionsToOffer]] where direction is not "stopped"".
```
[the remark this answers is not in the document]
alvestrand: Read the header extension spec. The full format is a=hdrext:<number>/<direction> <url>, with the <direction> defaulting to "sendrecv", so "a=hdrext:<number> <url>" being a shorthand.

sendrecv means that I can send it, you can send it, we both promise that we'll understand it.
sendonly means that I'm going to inflict it on you, but if you send it to me, I'll ignore it, so you're better off not wasting the bits.

index.html
```diff
@@ -122,6 +159,161 @@ <h2>Methods</h2>
...
+    </p>
+    <p class="note">
+      Since JSEP does not know about WebRTC internal slots, merging this change requires
+      more work on a JSEP revision.
+    </p>
+    <p>
+      In the algorithm to <a>set an SDP description</a>, add a bullet in the sections
```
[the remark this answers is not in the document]
alvestrand: Good point. It's missing (thus far) in the linked-references list.

index.html
alvestrand: Only negotiated extensions are returned from getParameters(). The URL field is read-only, so we throw if it is changed. Need to say that setParameters() throws if the length of the list is different, too.

index.html
alvestrand: Can we leave that to the header extensions spec? I really don't want to repeat everything it has to say about direction here.

index.html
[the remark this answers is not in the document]
henbos: OK got it, then direction makes sense.

index.html
henbos: That's fine I thought this was a new concept but it isn't.

index.html
```diff
@@ -60,8 +60,25 @@
...
     <p>
-      TODO(henbos): Fill in.
+      This document contains proposed extensions to the [[WEBRTC]] specification,
+      where it was thought inappropriate to incorporate them directly into that specification
+      at the point in time when they were written, but they were too small to warrant
+      writing a separate document.
+    </p>
+    <p>
+      If, at a later time, it is a good time to add them to the base specification, that
+      will be done. Deciding factors will include maturity of the extension, consensus on
```
jan-ivar: Too much passive language for my taste. I don't think we need to explain. How about just:

This document contains proposed extensions to the [[WEBRTC]] specification. These extensions may be added to the base specification should they reach maturity in implementation and gain consensus.

index.html
```diff
@@ -122,6 +159,165 @@ <h2>Methods</h2>
...
+      BUNDLE is in use.
+    </p>
+    <p>
+      Let [[\HeaderExtensionsNegotiated]] be an internal slot of the RTCRtpTransceiver,
+      initialized to an empty list.
+    </p>
+    <p>
+      In the algorithms for generating initial offers in JSEP section 5.1.1,
+      read "for each supported RTP header extension" as "For each RTP header extension
+      listed in [[\HeaderExtensionsToOffer]] where direction is not "stopped"".
```
henbos: The "for each supported RTP header extension" language is in JSEP 5.3.1 (Constructing an Offer - Initial Offers), not 5.1.1.

We need to say how the direction affects the SDP, a spec reference is enough. E.g. "The direction attribute of the a=extmap (RFC5285, Section 6) MUST reflect the direction attribute of the RTCRtpHeaderExtensionCapability".

What about for the answer? You did not modify the paragraph in JSEP 5.3.1 (Generating an Answer - Initial Answers). Is this an offer-only API or is it meant to control what is answered as well? Can I use this API to reject an RTP header extension the same way RTCRtpTransceiverDirection is used in both the offer AND answer in the case of RTCRtpTransceivers? Or will the answer always reply with the maximum direction it is capable of receiving, regardless if the endpoint wants to offer it?

index.html
```diff
@@ -122,6 +159,165 @@ <h2>Methods</h2>
...
+    </h3>
+    <p>
+      The <code><a href="https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver">RTCRtpTransceiver</a></code>
+      interface is defined in [[WEBRTC]]. This document extends that interface
+      by adding an additional method and attribute in order to control negotiation of
+      RTP header extensions.
+    </p>
+    <pre class="idl">
+partial dictionary RTCRtpHeaderExtensionCapability {
+  RTCRtpTransceiverDirection direction = "sendrecv";
```
jan-ivar: Needs a dictionary member section and explanation of `direction`, either here or in a follow up (I don't understand what it does, even from the discussion here).

index.html
```diff
@@ -122,6 +159,165 @@ <h2>Methods</h2>
...
+};
+      
+partial interface RTCRtpTransceiver {
+  void setOfferedRtpHeaderExtensions(
+    sequence&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer);
+  readonly attribute FrozenArray&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsNegotiated;
+};
+    </pre>
+    <p>
+      Let [[\HeaderExtensionsToOffer]] be an internal slot of the RTCRtpTransceiver,
```
jan-ivar: _"When [creating an RTCRtpTransceiver](https://w3c.github.io/webrtc-pc/#dfn-create-an-rtcrtptransceiver), add the following steps to the end:"_

index.html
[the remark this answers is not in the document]
jan-ivar: Can we add a direct link in the meantime?

index.html
```diff
@@ -122,6 +159,165 @@ <h2>Methods</h2>
...
+                </li>
+              </ol>
+            </li>
+            <li>
+              For each mandatory-to-implement extension <var>mext</var>, if <var>mext</var>
+<<<<<<< HEAD
+              is not present in <extlist>, throw an IllegalArgument error.
+=======
+              is not present in <var>extlist</var>, throw an IllegalArgument error.
+>>>>>>> 441d821224ddbc98724c985e9639cffe2e20344e
```
jan-ivar: meep meep

index.html
```diff
@@ -122,6 +159,165 @@ <h2>Methods</h2>
...
+            </li>
+            <li>
+              Set the [[\HeaderExtensionsToOffer]] to <var>extlist</var>.
+            </li>
```
henbos: "JSEP 5.2.2 Subsequent Offers" says:

> In addition, for each existing, non-recycled, non-rejected m= section in the new offer, the following adjustments are made based on the contents of the corresponding m= section in the current local or remote description, as appropriate:
> ...
> - The RTP header extensions MUST only include those that are present in the most recent answer.

If we want to override this behavior we must be clear that this is intentional. If we don't want to override this behavior, we should probably throw an exception if you attempt to use this API after [[\HeaderExtensionsNegotiated]] has a value.

index.html
henbos: Rebase issue

index.html
```diff
@@ -133,17 +329,17 @@ <h3>
     </p>
     <p>
       Let <code>RTCRtpReceiver</code> objects have a
-      <dfn>[[\PlayoutDelayHint]]</dfn> internal slot initially
+      <dfn>[[\PlayoutDelay]]</dfn> internal slot initially
```
jan-ivar: Unrelated to this PR?

index.html
```diff
@@ -122,6 +159,165 @@ <h2>Methods</h2>
...
+            sender should never send this extension.
+          </p>
+        </dd>
+      </dl>
+    </section>
+    <p>
+      The inclusion of a settable member of RTCRtpHeaderExtensionParameters
+      means that it is no longer a read-only member of RTCRtpParameters.
+      However, the "uri", "id" and "encrypted" members of
+      RTCRtpHeaderExtensionParameters are now read-only.
```
jan-ivar: I'd need more time to grok the overall reasons for this. Is there a short explainer I missed somewhere for what the use case is exactly? That would help me review this further.

index.html
```diff
@@ -122,6 +159,165 @@ <h2>Methods</h2>
...
+    <p class="note">
+      Since JSEP does not know about WebRTC internal slots, merging this change requires
+      more work on a JSEP revision.
+    </p>
+    <p>
+      In the algorithm to <a>set an SDP description</a>, add a bullet in the sections
+      for processing answer and pranswer (bullet 3.2.8.1.7 and 3.2.8.2.6):
+      <ul>
+        <li>Set the transceiver's [[\HeaderExtensionsNegotiated]] slot to the list of
+          RTP header extensins from the answer.
```
henbos: If the transceiver is created because of a remote offer I think we need to set [[\HeaderExtensionsToOffer]] to match the offer, or to say "Set [[\HeaderExtensionsToOffer]] to [[\HeaderExtensionsNegotiated]]" here. Otherwise we will break JSEP rules, which says that any follow-up offers must have the same RTP header extensions as the previous answer.

henbos: Actually the way JSEP is written I think HeaderExtensionsToOffer would simply be ignored in follow-up offers if the m= section has previously been negotiated.

alvestrand:
Note on RID: The default browser can send rid, but can't receive it. So the negotiation control needs to say that the browser may restrict what directions can be set for specific extensions - not just say "mandatory must be sendrecv".

alvestrand:
Add a comment that getParameters() on the receiver will not return "enabled".

alvestrand:
Should this API only be callable before initial negotiation, or should we modify more JSEP sections (createAnswer, subsequent create*)

henbos:
Please re-upload this PR (looks like merge changes are part of this PR), address comments, and ask for more review.

alvestrand:
@henbos looks as if you're not looking at the 3:55 PM version - I think I got the unresolved diff-mark out of the file in that upload.

index.html
alvestrand: Somehow github is showing diff against the wrong version. That change has been merged to head according to github.io

detetivepro12: Junior Oliveira

index.html
alvestrand: This particular clause just notes that the headerExtensions member of RTCRtpParameters used to be read-only, and it is no longer read-only. The sentence needs to be changed to "remain read-only", since the base spec already stated that they were read-only.

alvestrand:
A rebase-and-track dance brought the diff in line with what was expected (no more PlayoutDelayHint in the diffs).

henbos:
Ack, let us know when you want us to take another look

alvestrand:
Got commentary from the guy who's implementing it - leading to some proposed changes:
- Expose what's currently set to be sent
- That should also expose the default before calling the setter, which allows us to include extensions that aren't on by default as having direction "stopped"
- it's probably better to expose setting via a setter method than by modifying the attribute, since the usual tactic is to configure the whole thing in one go rather than piecemeal.

index.html
alvestrand: Tried a different approach, saying that setting [[HeaderExtensionsToOffer]] can restrict the list of offered extensions, but cannot extend it. That should provide interoperability with implementations that depend on the list never growing (if any).

alvestrand:
Please take another look.

index.html
```diff
@@ -122,6 +177,196 @@ <h2>Methods</h2>
...
+      an appropriate "a=extmap" line with "direction" set according to the rules of [[RFC5285]]
+      section 6, considering the direction in [[\HeaderExtensionsToOffer]] to indicate the
+      answerer's desired usage". This preserves the property that the set of extensions
+      may not grow with subsequent offers.
+    </p>
+    <p>
+      In the algorithm for generating initial answers in [[RTCWEB-JSEP]] section 5.3.1, replace "For
+      each supported RTP header extension that is present in the offer" with "For each
+      supported RTP header extension that is present in the offer and is also present in
+      [[\HeaderExtensionsToOffer]] with a direction different from "stopped"".
```
henbos: This sentence makes it sound like while we can reject an extension by stopping it, the direction cannot be affected.

For example, if "sendrecv" is offered and I use the API to set it to "sendonly", the generated answer will still say "sendrecv".
This is not the case for transceiver directions: e.g. "sendrecv" can be offered and "sendonly" or "recvonly" can be replied.

Is this on purpose?

index.html
```diff
@@ -122,6 +177,196 @@ <h2>Methods</h2>
...
+      by adding an additional method and attribute in order to control negotiation of
+      RTP header extensions.
+    </p>
+    <pre class="idl">
+partial dictionary RTCRtpHeaderExtensionCapability {
+  RTCRtpTransceiverDirection direction = "sendrecv";
+};
+      
+partial interface RTCRtpTransceiver {
+  void setOfferedRtpHeaderExtensions(
```
henbos: The method and internal slot is used for both createOffer() and createAnswer() but the names (setOfferedRtpHeaderExtensions, headerExtensionsToOffer, headerExtensionsNegotiated) sound like they are special to the offer.

Suggestion (borrowing the rules of direction/currentDirection): writeable headerExtensions and readonly currentHeaderExtensions.

index.html
```diff
@@ -122,6 +177,196 @@ <h2>Methods</h2>
...
+                <li>If <var>ext</var>'s "direction" field is not "sendrecv", and "uri"
+                  indicates a mandatory-to-use attribute that is required to be both
+                  sent and received, throw an IllegalArgument error.
+                </li>
+                <li>if <var>ext</var>'s "direction" field is "stopped", and "uri" indicates
+                  a mandatory-to-implement extension, throw an IllegalArgument error.
+                </li>
+                <li>
+                  Set the "direction" field of <var>entry</var> to <var>ext</var>'s
+                  "direction" field.
```
henbos: This algorithm could modify one entry and throw an exception at the next entry. That would be a bug.

Make a copy of all the entries, and only if we manage to iterate all ext:s without throwing to we set [[\HeaderExtensionsToOffer]] as the last step before returning.

index.html
```diff
@@ -122,6 +177,196 @@ <h2>Methods</h2>
...
+      </ul>
+    </p>
+    <section>
+      <h2>Methods</h2>
+      <dl>
+        <dt>setOfferedRtpHeaderExtensions</dt>
+        <dd>
+          Execute the following steps:
+          <ol>
+            <li>Let <var>extlist</var> be the list of extensions in the argument.</li>
```
henbos: Add a step that says that if [[\HeaderExtensionsNegotiated]] is non-empty and we are attempting to set the direction of a stopped entry that is not presently negotiated to a value other than "stopped", throw an exception.

Since we are not allowed to add to the set of negotiated extensions after the initial negotiation, we should fail rather than silently ignore that extension in the follow-up offer.

index.html
```diff
@@ -122,6 +177,196 @@ <h2>Methods</h2>
...
+    </section>
+    <section>
+      <h2>Modifications to existing procedures</h2>
+      <p>
+        In the <a>final steps to create an offer</a>, add a step:
+      </p>
+      <ol>
+        <li>If the [[\HeaderExtensionsToOffer]] internal slot is set, for
+          each extension in [[\HeaderExtensionsToOffer]], add an appropriate
+          "a=hdrext" line.
```
henbos: Is "a=hdrext" supposed to say "a=extmap"?

index.html
```diff
@@ -122,6 +177,196 @@ <h2>Methods</h2>
...
+        </li>
+      </ol>
+      <p>
+        In the <a>set an RTCSessionDescription</a> algorithm, add a step
+        right after the step that sets transceiver.[[\Sender]].[[\SendCodecs]],
+        saying "For each transciever, set [[\HeaderExtensionsNegotiated]] to
+        an empty list, and then
+        for each "a=hdrext" line, add an appropriate
+        RTCRtpHeaderExtensionCapability to the list.
```
henbos: "a=extmap"?

index.html
```diff
@@ -122,6 +177,196 @@ <h2>Methods</h2>
...
+        <dd>Returns the value of the [[\HeaderExtensionsToOffer]] slot.</dd>
+      <dl>
+        <dt>headerExtensionsNegotiated, readonly</dt>
+        <dd>Returns the value of the [[\HeaderExtensionsNegotiated]] slot.</dd>
+      </dl>
+    </section>
+    <section>
+      <h2>Modifications to existing procedures</h2>
+      <p>
+        In the <a>final steps to create an offer</a>, add a step:
```
henbos: Do we need to do the same in "final steps to create an answer"? Also is this repeating the same thing as the earlier paragraphs that was referring to JSEP or is this doing something new too?

index.html
```diff
@@ -122,6 +177,196 @@ <h2>Methods</h2>
...
+      partial dictionary RTCRtpHeaderExtensionParameters {
+        boolean enabled;
+      };
+    </pre>
+    <section>
+      <h2>Attributes</h2>
+      <dl>
+        <dt>enabled, of type boolean</dt>
+        <dd>
+          <p>When returned from getParameters() on an RTPSender, "enabled" is true
```
henbos: RTCRtpSender

index.html
```diff
@@ -122,6 +177,196 @@ <h2>Methods</h2>
...
+            configured to send this header extension when appropriate. If the
+            attribute is missing, it means that it is enabled and cannot be disabled.
+          </p>
+          <p>
+            When passed to setParameters(), setting "enabled" to true instructs the
+            RTP sender to send this header extension when appropriate. Setting "enabled"
+            to false instructs the RTP sender to never send this extension.
+          </p>
+          <p>
+            When calling getParameters() on an RTPReceiver, the "enabled" member will
```
henbos: RTCRtpReceiver

index.html
```diff
@@ -122,6 +177,196 @@ <h2>Methods</h2>
...
+            always be missing, indicating that there is no control surface here.
+          </p>
+        </dd>
+      </dl>
+      <p class="note">
+        The list of extensions returned from getParameters() on a sender will only include
+        the extensions that have been negotiated for sending. Changing the
+        list, apart from setting the "enabled" member, is not permitted.
```
henbos: nit: Maybe make this part of the normative steps? It's testable and it's not clear without the note.

index.html
```diff
@@ -122,6 +177,196 @@ <h2>Methods</h2>
...
+            attribute is missing, it means that it is enabled and cannot be disabled.
+          </p>
+          <p>
+            When passed to setParameters(), setting "enabled" to true instructs the
+            RTP sender to send this header extension when appropriate. Setting "enabled"
+            to false instructs the RTP sender to never send this extension.
+          </p>
+          <p>
+            When calling getParameters() on an RTPReceiver, the "enabled" member will
+            always be missing, indicating that there is no control surface here.
```
henbos: nit: "indicating that there is no control surface here" is not needed, because RTCRtpReceiver.setParameters() is not a method that exists, so this is true for anything that it returns. Just say it's always missing.

index.html
```diff
@@ -122,6 +177,196 @@ <h2>Methods</h2>
...
+  readonly attribute FrozenArray&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer;
+  readonly attribute FrozenArray&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsNegotiated;
+};
+    </pre>
+    <p>
+      Let [[\HeaderExtensionsToOffer]] be an internal slot of the RTCRtpTransceiver,
+      initialized to the platform-specific list of implemented RTP header extensions.
+      The direction
+      attribute for all extensions that are mandatory to use MUST be initialized to
+      an appropriate value other than "stopped". The direction attribute for extensions
```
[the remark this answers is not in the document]
alvestrand: I'm not clear on whether getCapabilities() should return direction at all; it is able to do so because we used "partial dictionary", but that doesn't mean that we have to.
I think the Principle of Least Astonishment says that sender.getCapabilities() should return the same thing as "addTrancsceiver().headerExtensionsToOffer()", filtered to remove header extensions that can't be set on a sender (if any). Returning differing values seems like a gotcha that we don't need to make.

index.html
```diff
@@ -122,6 +177,196 @@ <h2>Methods</h2>
...
+      that will not be offered by default in an initial offer MUST be initialized to
+      "stopped".
```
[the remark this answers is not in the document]
alvestrand: Yes, in this API, that's the way to do it.

index.html
```diff
@@ -122,6 +177,196 @@ <h2>Methods</h2>
...
+    <p class="note">The list of header extensions that MUST/SHOULD be supported is listed in
+      [[RTCWEB-RTP]], section 5.2. The "mid" extension is mandatory to use when
+      BUNDLE is in use, per [[BUNDLE]] section 9.1.
+    </p>
+    <p>
+      Let [[\HeaderExtensionsNegotiated]] be an internal slot of the RTCRtpTransceiver,
+      initialized to an empty list.
+    </p>
+    <p>
+      In the algorithms for generating initial offers in [[RTCWEB-JSEP]] section 5.1.1,
```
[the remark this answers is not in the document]
alvestrand: Correct. Corrected.

index.html
```diff
@@ -122,6 +177,196 @@ <h2>Methods</h2>
...
+        <dt>setOfferedRtpHeaderExtensions</dt>
+        <dd>
+          Execute the following steps:
+          <ol>
+            <li>Let <var>extlist</var> be the list of extensions in the argument.</li>
+            <li>
+              For each extension <var>ext</var> in <var>extlist</var>, do the following:
+              <ol>
+                <li>If <var>ext</var>'s "uri" value is not present in
+                  [[\HeaderExtensionsToOffer]], throw a NotSupported error.
```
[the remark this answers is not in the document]
alvestrand: Good idea, added the TypeError step for missing uri.
NotSupportedError is for when you pass in an URI that the platform doesn't know about.

index.html
```diff
@@ -122,6 +177,196 @@ <h2>Methods</h2>
...
+                </li>
+                <li>
+                  Let <var>entry</var> be the entry with the same "uri" value in
+                  [[\HeaderExtensionsToOffer]].
```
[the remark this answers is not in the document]
alvestrand: Throwing is already done in the previous step.

index.html
alvestrand: I deleted this. I think I duplicated these mods by accident.

index.html
alvestrand: Done.

index.html
alvestrand: Done

index.html
alvestrand: Removed the "class=note".

index.html
alvestrand: Done.

index.html
```diff
@@ -122,6 +177,185 @@ <h2>Methods</h2>
...
+        an appropriate "a=extmap" line with "direction" set according to the rules of [[RFC5285]]
+        section 6, considering the direction in [[\HeaderExtensionsToOffer]] to indicate the
+        answerer's desired usage". This preserves the property that the set of extensions
+        may not grow with subsequent offers.
+      </p>
+      <p>
+        In the algorithm for generating initial answers in [[RTCWEB-JSEP]] section 5.3.1, replace "For
+        each supported RTP header extension that is present in the offer" with "For each
+        supported RTP header extension that is present in the offer and is also present in
+        [[\HeaderExtensionsToOffer]] with a direction different from "stopped"".
```
henbos: If the transceiver was created by setting the offer, [[\HeaderExtensionsToOffer]] would have its initial value of whatever the default is. Which means that if the offer contains a non-default header extension that we support, then by default we would reject it.

For transceiver that were created as a result of SRD(offer), might it make more more sense to initialize [[\HeaderExtensionsToOffer]] to the subset of what's on offer that we are capable of accepting?

This has several benefits:
1) The answerer does not have to figure out which header extensions to add in the most common case of accepting whatever is on offer.
2) This follows a similar pattern as transceiver.direction/currentDirection.
3) If between offer and answer, the answerer wants to know what was on offer, it can just check transceiver.headerExtensionsToOffer, because it reflects the offer. No SDP parsing needed.

index.html
```diff
@@ -122,6 +177,185 @@ <h2>Methods</h2>
...
+      </p>
```
henbos: Same comment as in earlier commit: What if what's offered is "sendrecv" but headerExtensionsToOffer is "recvonly"? We should reply with the appropriate direction, not JUST filter out stopped ones.

index.html
```diff
@@ -122,6 +177,185 @@ <h2>Methods</h2>
...
+      <dfn>RTCRtpTransceiver</dfn> extensions
+    </h3>
+    <p>
+      The <code><a href="https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver">RTCRtpTransceiver</a></code>
+      interface is defined in [[WEBRTC]]. This document extends that interface
+      by adding an additional method and attribute in order to control negotiation of
+      RTP header extensions.
+    </p>
+    <pre class="idl">
+partial dictionary RTCRtpHeaderExtensionCapability {
```
henbos: Can you add a sentence somewhere saying that this attribute is not set for getCapabilities()?

index.html
```diff
@@ -122,6 +177,185 @@ <h2>Methods</h2>
...
+            attribute is missing, it means that it is enabled and cannot be disabled.
+          </p>
+          <p>
+            When passed to setParameters(), setting "enabled" to true instructs the
+            RTP sender to send this header extension when appropriate. Setting "enabled"
+            to false instructs the RTP sender to never send this extension.
+          </p>
+          <p>
+            When calling getParameters() on an RTCRTPReceiver, the "enabled" member will
+            always be missing.
```
henbos: RTCRtpReceiver

index.html
```diff
@@ -122,6 +177,185 @@ <h2>Methods</h2>
...
+      partial dictionary RTCRtpHeaderExtensionParameters {
+        boolean enabled;
+      };
+    </pre>
+    <section>
+      <h2>Attributes</h2>
+      <dl>
+        <dt>enabled, of type boolean</dt>
+        <dd>
+          <p>When returned from getParameters() on an RTCRTPSender, "enabled" is true
```
henbos: RTCRtpSender

henbos:
I'm happy to merge and iterate in follow-up PRs if we can all agree on the name of the attribute in this PR. Me arguing for "headerExtensions" rather than "headerExtensionsToOffer".

index.html
```diff
@@ -122,6 +177,185 @@ <h2>Methods</h2>
...
+      RTP header extensions.
+    </p>
+    <pre class="idl">
+partial dictionary RTCRtpHeaderExtensionCapability {
+  RTCRtpTransceiverDirection direction = "sendrecv";
+};
+      
+partial interface RTCRtpTransceiver {
+  void setOfferedRtpHeaderExtensions(
+sequence&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer);
```
alvestrand: Rename to headerExtensionsToNegotiate.

henbos:
Also headerExtensionsToNegotiate should be initialized to what is on offer if the transceiver was created by the offer. A subset of the what's on offer that the endpoint is capable of accepting

alvestrand:
I don't think modifying headerExtensionsToNegotiate on incoming offers is the Right Thing. If so, we need to change the setOfferedRtpHeaderExtensions algorithm to permit-but-ignore URIs that are supported but are not on offer.
We already apply the filtering function in "generate an answer" to only reply with stuff that is also present in the offer; keeping "toNegotiate" as "what I set" makes life simpler.

alvestrand:
@henbos you need to remove your "changes requested" before I can submit.

PR #37, opened 2020-05-12
Add generateKeyframe() API

alvestrand:
Fix for Issue https://github.com/w3c/webrtc-extensions/issues/72

This was the simplest API I could think of.
The use of the operations chain (which I haven't been able to cite correctly) is to make it not racy with replaceTrack(), which is also on the operations chain.

I'd like to have comments from hbos and jib before proposing this to the WG.

***
<a href="https://pr-preview.s3.amazonaws.com/alvestrand/webrtc-extensions/pull/37.html" title="Last updated on Nov 15, 2021, 1:39 PM UTC (7878543)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-extensions/37/c5e02da...alvestrand:7878543.html" title="Last updated on Nov 15, 2021, 1:39 PM UTC (7878543)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -401,7 +401,7 @@ sequence&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer);
       extends that interface by adding an additional method.
     </p>
     <pre class="idl">partial interface RTCRtpSender {
-      Promise&lt;unsigned long> generateKeyFrame();
+      Promise&lt;unsigned long> generateKeyFrame(optional sequence &lt;DOMString> rids);
       };</pre>
     <section>
       <h2>Methods</h2>
@@ -425,10 +425,14 @@ sequence&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer);
               <li>
                 Return  <var>p</var>, and let it be the result of
                 [= chaining =] the following steps to
-                    <var>connection</var>'s [= operations chain =].
+                <var>connection</var>'s [= operations chain =].
               </li>
               <li>If the sender's 'kind' is not "video", reject <var>p</var>
                 with {{NotAllowedError}}.
+              </li>
+              <li>If the <code>rids</code> argument is provided, check whether
+                  each of the RID values in the sequence are valid. If any RID
+                  values are invalid, reject <var>p</var> with {{NotAllowedError}}.
               </li>
               <li>
                 If the sender's track is not set, if the track is ended,
@@ -436,15 +440,19 @@ sequence&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer);
                 {{InvalidStateError}}.
               </li>
               <li>
-                Instruct the encoder of the track to produce a keyframe.
+                If the <code>rids</code> argument is provided, instruct the
+                encoder of the track to generate keyframe(s) for each of the
+                simulcast streams represented by <code>rids</code>. If
+                <code>rids</code> is not provided, generate keyframe(s) for
+                every simulcast stream.
               </li>
               <li>
-                If the production of a keyframe fails, reject <var>p</var>
+                If generating keyframe(s) fails, reject <var>p</var>
                 with {{OperationError}}.
               <li>
-                When the encoder has produced a keyframe, and it has been
-                assigned an RTP timestamp, resolve <var>p</var> with this
-                timestamp.
+                When the encoder has generated keyframe(s) and has
+                assigned an RTP timestamp, resolve <var>p</var> with
+                the timestamp.
               </li>
             </ul>
           </dd>
@@ -462,7 +470,6 @@ sequence&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer);
       </pre>
     </p>
   </section>
-
   <section id="rtcrtpreceiver-interface">
     <h3>
       {{RTCRtpReceiver}} extensions
--- a/index.html
+++ b/index.html
@@ -392,6 +392,77 @@ sequence&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer);
       with an InvalidModificationError.
     </p>
   </section>
+  <section id="recrtpsender-interface">
+    <h3>
+      {{RTCRtpSender}} extensions
+    </h3>
+    <p>
+      The {{RTCRtpSender}} interface is defined in [[WEBRTC]]. This document
+      extends that interface by adding an additional method.
+    </p>
+    <pre class="idl">partial interface RTCRtpSender {
+      Promise&lt;unsigned long> requestKeyFrame();
+      };</pre>
+    <section>
+      <h2>Methods</h2>
+      <p>
+        <dl data-link-for="RTCRtpSender" class="methods">
+          <dt>
+            <dfn data-idl><code>requestKeyFrame</code></dfn></dt>
+          <dd>
+            <p>
+              Requests a keyframe. When called, the UA MUST perform
+              the following steps:
+            </p>
+            <ul>
+              <li>
+                Let <var>p</var> be a new promise.
+              </li>
+              <li>
+                Let <var>connection</var> be the sender''s associated
+                {{RTCPeerConnection}}.
+              </li>
+              <li>
+                Return  <var>p</var>, and let it be the result of
+                [= chaining =] the following steps to
+                    <var>connection</var>'s [= operations chain =].
+              </li>
+              <li>If the sender''s 'kind' is not "video", reject <var>p</var>
+                with {{NotAllowedError}}.
+              </li>
+              <li>
+                If the sender''s track is not set, if the track is ended,
+                or if the track is muted, reject <var>p</var> with
+                {{InvalidStateError}}.
+              </li>
+              <li>
+                Instruct the encoder of the track to produce a keyframe.
+              </li>
+              <li>
+                If the production of a keyframe fails, reject <var>p</var>
+                with {{OperationError}}.
+              <li>
+                When the encoder has produced a keyframe, and it has been
+                assigned an RTP timestamp, resolve <var>p</var> with this
+                timestamp.
+              </li>
+            </ul>
+          </dd>
+        </dl>
+      </p>
+    </section>
+    <p>
+      This call can be emulated by making the following calls,
+      except that the outgoing RTP timestamp is not recorded, and
+      the encoder may choose to encode an extra black frame:
+      <pre>
+        let track = sender.track;
+        await sender.replaceTrack(null);
+        await sender.replaceTrack(track);
+      </pre>
+    </p>
+  </section>
+
   <section id="rtcrtpreceiver-interface">
     <h3>
       {{RTCRtpReceiver}} extensions
--- a/index.html
+++ b/index.html
@@ -419,7 +419,7 @@ sequence&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer);
                 Let <var>p</var> be a new promise.
               </li>
               <li>
-                Let <var>connection</var> be the sender''s associated
+                Let <var>connection</var> be the sender's associated
                 {{RTCPeerConnection}}.
               </li>
               <li>
--- a/index.html
+++ b/index.html
@@ -452,9 +452,9 @@ sequence&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer);
       </p>
     </section>
     <p>
-      This call can be emulated by making the following calls,
-      except that the outgoing RTP timestamp is not recorded, and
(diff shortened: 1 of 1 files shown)


murillo128:
I like the API!

How does it behave with simulcast and SVC (K-SVC)? Can an key frame be requested for particular layer/stream (like the RTCP LRR)?

alvestrand:
Good question. I think it needs a layer argument (with default being "all layers").

alvestrand:
@aboba is there a way in some SVC spec to refer consistently to a specific layer?
Does the question even make sense?
For simulcast, indexing the layers in Encodings should be sufficient.

fippo:
@alvestrand related to e2ee stuff?
Please note the opinion of the google team on PLIs/keyframes in relation to simulcast and per-stream keyframes, see https://bugs.chromium.org/p/webrtc/issues/detail?id=10107 (also :wave: @ibc)

IIRC a SRD/SLD also triggers a keyframe (or rather: encoder reset)

ibc:
What is this API needed for?

aboba:
@alvestrand The [FIR message](https://tools.ietf.org/html/rfc5104#section-4.3.1.1) utilizes SSRC(s) to indicate the target(s) of the request. So the FIR is assumed to apply to all layers on the target SSRC, [as well as associated SSRCs in the case of layered codecs](https://tools.ietf.org/html/rfc8082#section-4). The [LRR](https://tools.ietf.org/html/draft-ietf-avtext-lrr) utilizes an SSRC, but also includes other fields such TTID (the target temporal layer), TLID (the target spatial layer), etc.

Note the updated definition of "decoder refresh point" in [RFC 8082](https://tools.ietf.org/html/rfc8082#page-4).

aboba:
@murillo128 How it behaves will depend on what message gets generated as a result (FIR or LRR). For an FIR, the encoder(s) associated with the target SSRCs will generate a refresh point. If a single encoder is generating simulcast, this can result in **all** of the simulcast streams generating a key frame, as described [here.](https://aomediacodec.github.io/av1-rtp-spec/#81--full-intra-request-fir). LRR behavior is described [here. ](https://aomediacodec.github.io/av1-rtp-spec/#82--layer-refresh-request-lrr)

youennf:
It would be good to understand the use case.
For instance, would applications like to control the maximum time between key frames an encoder uses, with two extreme cases being key frame only and no key frame except in case of FIR?

A few additional questions:
- Would it also be useful for some audio codecs?
- Is it WebRTC specific or is there a use case for MediaRecorder as well?
- What does it mean for muted content? I guess the next frame would be a key frame, which might require waiting for 1 sec, so the promise would resolve after 1 sec. Is it correct?
- What should be the behaviour if this method is called 10 times synchronously? Should there be one key frame or should the next 10 frames be key frames?
- Is it an issue to be able to compute how long it will take to generate a key frame (say I call this API on a canvas track and after some time, I generate a canvas frame and compute the time between the canvas frame generation and the promise resolution)?
- This is using the operation chain, probably to keep consistency with the replaceTrack trick. Is that so? Is there any other reason? Is there a future need to control exactly which frame would get encoded as a key frame?

index.html
```diff
@@ -392,6 +392,77 @@ <h2>Attributes</h2>
...
+            <p>
+              Requests a keyframe. When called, the UA MUST perform
+              the following steps:
+            </p>
+            <ul>
+              <li>
+                Let <var>p</var> be a new promise.
+              </li>
+              <li>
+                Let <var>connection</var> be the sender''s associated
```
alvestrand, line 422: delete double quote

alvestrand:
@youennf the use case that came up this time is to make sure a keyframe is sent each time you change the encryption key on a group call.

Perhaps this is better called "produceKeyframe"?

aboba:
Didn't realize this was for the sender, not the receiver.  So it's not really a request, it's more like "generateKeyFrame".

youennf:
I see, so it might make sense for audio if codecs use some prediction. Maybe the promise should not reject then?

An alternative solution is for the e2e JS transform module to state that it needs a key frame as part of the additional frame metadata given to the packetizer for each processed frame. That might be also useful for a non JS e2e module for which the key changes might not be visible to the script.

ibc:
Let me insist, please. What is this API needed for? And if the answer is e2ee, why?

murillo128:
the e2ee key is exchanged in a different path than sfu signaling, so it may
be updated on the sender shortly after the new participant has joined and
the sfu jas already requested the key frame.

Updating the key on a key frame makes a lot of sense to me.

Anyway, I am always in favor of more APIs that allow lower level control of
the underlaying objects.

Do you foresee any way this API could cause any issue to existing code?

ibc:
> Do you foresee any way this API could cause any issue to existing code?

No, but neither I can figure out how your response explains what this API is needed for :)

murillo128:
Forward and Post-Compromise Security requires that the e2ee keys are updated anytime a participant joins/leave the call.
https://tools.ietf.org/html/draft-ietf-mls-architecture-04#section-3.2.2.1

The key exchange happens async and on a different path than the webrtc signaling and media (by definition). So it may happen that when a new participant joins the SFU side request a key frame and the sender generates the e2ee encrypted frame with a key not known by the receiver, so it will be discarded. When the sender updates his sending key with the new key, it will send it in a non-key frame frame, so the receiver will be able to decrypt it, but not decode it.

Receiver will re-request an key frame then, but due to sender and sfu policies, that new key frame could take some time to be generated.

It the sender sends a key frame when the new e2ee key is in use, the time required for the new participant to display the video is minimized.

Also, there are several optimizations (specially on authenticating and signing) that can be done on key frames, so it is always interesting to be able to control the key frame generation from the e2ee side.

Hope it is clearer now.

index.html
```diff
@@ -392,6 +392,77 @@ <h2>Attributes</h2>
       with an InvalidModificationError.
     </p>
   </section>
+  <section id="recrtpsender-interface">
```
fippo, line 395: s/rectpsender/rtpsender/

index.html
```diff
@@ -392,6 +392,77 @@ <h2>Attributes</h2>
...
+              <li>
+                Return  <var>p</var>, and let it be the result of
+                [= chaining =] the following steps to
+                    <var>connection</var>'s [= operations chain =].
+              </li>
+              <li>If the sender''s 'kind' is not "video", reject <var>p</var>
+                with {{NotAllowedError}}.
+              </li>
+              <li>
+                If the sender''s track is not set, if the track is ended,
```
fippo, line 434: delete double quote

fippo:
In particular in the context of insertable streams I don't care about when the next keyframe happens, no? I could request one to be generated and would rotate the key at the next keyframe that the transform sees. Now if there are two changes to the e2ee key in rapid succession... then what?

I don't think resolving with a single timestamp works for simulcast where we have three different timestamp sequences.

fippo:
> Didn't realize this was for the sender, not the receiver. So it's not really a request, it's more like "generateKeyFrame".

Do we need a requestKeyFrame on the receiver? Mainly to handle decrypt errors.

youennf:
Given the use case, this API might better fit with insertable streams.
To reduce race issues, this API should be tied as close as possible to the encoder, the ReadableStream being the closest we have.

> Do we need a requestKeyFrame on the receiver? Mainly to handle decrypt errors.

Seems to make sense, for metadata as well in theory. That is a bit more tricky since this API is trying to influence the encoder of the other side, communication channel being potentially lossy.

youennf:
@alverstrand, given the usecase is insertable streams, should we move this issue to webrtc-insertable-streams repo?

solmaks:
@alvestrand , I think we should resurrect this work. RTCRtpSender::replaceTrack workaround is not working reliably (seems like a race), and other workarounds (e.g. scaleResolutionBy toggle) are fundamentally flawed too. It seems that time is right for adding a first party API to request key frames on sender.

youennf:
@solmaks, can you share your usecase? Is it tied to using webrtc encoded transforms?
Safari has experimental support for this requestKeyFrame API as part of its insertable streams API experiment (in RTCRtpScriptTransformer) since there are known use cases there.
Using an identity transform just for getting access to this API would be cumbersome and might prove that this is not the right place (or only place) for such an API.

solmaks:
Actually it is not. Scenario is dead simple - adjust send stream bitrate and trigger key frame right after. We do have a mechanism for adjusting bitrate of individual encodings, per encoding key frame triggering is missing fully. It would be great if we could avoid bundling KF API with insertable streams or any other mechanisms involving heavy per-packet processing, as the latter has impact on power and performance.

youennf:
> adjust send stream bitrate and trigger key frame right after.

I see. In the case of simulcast, you would like to trigger a key frame for one encoding but not the others.
The closest API surface we have here is RTCRtpSendParameters.encodings.
I would have hoped that, when we go from active=false to active=true, video encoders would start with a key frame.
Maybe this is something we could agree on.

> It would be great if we could avoid bundling KF API with insertable streams or any other mechanisms involving heavy per-packet processing, as the latter has impact on power and performance.

I agree it would be cumbersome to use, not sure about power impact if frames are not mutated. Cost might be reduced to two thread hops.

solmaks:
> I see. In the case of simulcast, you would like to trigger a key frame for one encoding but not the others.
> The closest API surface we have here is RTCRtpSendParameters.encodings.

Yes, this is what I was thinking of as well. This new parameter might need a special treatment though, as it is transient in nature. I.e. configuring it on an encoding and reading it back would not make much sense.

> I would have hoped that, when we go from active=false to active=true, video encoders would start with a key frame.
> Maybe this is something we could agree on.

Unfortunately active=false/true might have undesirable side-effects, such as actually suspending video stream prior to generating a key frame. Idea is to be able to generate a key frame on a continuous stream. Let me reiterate the scenario, update send bitrate on an already active stream, and request KF to be generated with new bitrate constraint being effective.

youennf:
> actually suspending video stream prior to generating a key frame
The sender could probably stop sending during one event loop, so the suspension might not be that big a deal in practice.
This would need experimentation.
Also, generating key frame is probably not something that should happen often.

solmaks:
@youennf , I am not sure how experimentation can help. Observing some behavior in current implementation does not mean it will not change or break in subsequent version. Unless of course we explicitly want to update specification, demanding for active=true/false toggle in a single even loop should not result in stream getting suspended (which I hope we are not willing to do).

solmaks:
> Also, generating key frame is probably not something that should happen often.

API needs to be robust and produce defined behavior regardless of the frequency it is called. Be it once per minute or once per second. My scenario does not require sub-second processing. However, 1s+ is something that is pretty much required.

youennf:
I quickly looked at active usage in WebRTC backend. It seems it currently triggers recreation of the whole encoder, so it probably triggers keyframe (good for a short term fix), but for all simulcasted encoders (not great...).

youennf:
@solmaks, I filed https://github.com/w3c/webrtc-extensions/issues/72 to keep track of this issue given this thread is related to a PR.

solmaks:
@youennf , @alvestrand , any recommendation on how we can proceed with the proposed extension?

youennf:
https://github.com/w3c/webrtc-extensions/issues/72 is probably the natural place to continue the discussion.
I think we would need to build a proposal and present it at next WebRTC interim.
Would you be able to carry that forward?
A look at how it could be implemented in browsers in the simulcast case would also be valuable.

fippo:
> In the case of simulcast, you would like to trigger a key frame for one encoding but not the others.

No. Simulcast is (at least in chrome) not able to do keyframes for individual layers, see https://bugs.chromium.org/p/webrtc/issues/detail?id=10107

solmaks:
@fippo, I think it is OK to have an API enabling the flexibility to control individual layers, and individual user agents to choose not to do this optimization.

youennf:
> I think it is OK to have an API enabling the flexibility to control individual layers, and individual user agents to choose not to do this optimization.

That would be sad if different browsers have different behaviours for this API.
This seems like a good topic for next interim.

aboba (changes requested):
I'd suggest adding an optional argument to specify the streams on which a refresh point should be generated. Inclusion of streams would be interpretted the same as reception of an FIR (which can specify multiple SSRCs).  Since RIDs are exposed and SSRCs may not be, it seems like the best way to accomplish this would be by providing an optional argument containing a sequence of RIDs.  If the optional argument is not provided, then it would apply to all streams sent by the `RTCRtpSender`.

solmaks:
Good idea, @aboba . This seems to be a better place for it, as opposed to having it on top of encoding configuration.

aboba:
Closed due to merger of [PR 125](https://github.com/w3c/webrtc-encoded-transform/pull/125).

PR #56, opened 2020-11-24
A couple of minor Respec fixes

dontcallmedom:


henbos:
Please take a look @dontcallmedom there's an error

alvestrand:
seems that the error is in the check that the CL adds..... probably an old error.

dontcallmedom:
error fixed, and quite a few additional changes brought in in the process

Issue #64, opened 2021-01-21
Investigate the possibility to interact with data channel in workers


youennf:
We could try to allow processing of data channel messages in workers, without trying to solve the bigger issue of creating a data channel in a worker.
That would solve some known use cases like Zoom or streaming

zenhack:
This is possibly related to some things I've been thinking about lately: in particular, I'd like to be able to transfer various webrtc-related objects between iframes in a page using postMessage() -- it seems like making RTCDataChannel [Transferrable][1] would solve this for workers as well.

(For my own use case I'm also interested in doing this with media streams and other objects)

[1]: https://developer.mozilla.org/en-US/docs/Web/API/Transferable

youennf:
@zenhack, could you share more of your use case?
Plan is to present something along the lines of https://docs.google.com/presentation/d/1crumgYj4eHkjo04faLktPTg0QoYJhTFoosEBudfJBuw/present?slide=id.g7957cd038b_9_43 and I could add your usecase there as well.

As of transferring MediaStreamTrack objects, this might best be tracked in a separate issue in https://github.com/w3c/mediacapture-extensions.
This might help other use cases such as use cases discussed in https://github.com/w3c/mediacapture-screen-share/issues/158

zenhack:
Sure!

This is in the context of [Sandstorm][1]. Basically, it's a platform for self-hosting web apps, and we don't want apps to be able to "phone home" to app developers without the consent of the user -- so on the server side apps are containerized and not given network access, and in the browser we use Content-Security-Policy to block most ways an app could contact the outside world (there's currently no way to do this for webrtc, which I'm trying to help solve in https://github.com/w3c/webappsec-csp/pull/457).

We also run the app's page itself from a separate origin than the Sandstorm UI, inside of an iframe, with the Sandstorm UI around it. You can get a feel for it at <https://demo.sandstorm.io>. The iframe has the unfortunate consequence that apps can't request access to mic & camera directly.

So, I'd like to provide a way through which apps could request this access (and also controlled use of the networking portions of webrtc) from Sandstorm itself via a postMessage() API; we already use this mechanism for requesting access to server-provided resources, so it would be a natural extension: https://docs.sandstorm.io/en/latest/developing/powerbox/

[...]

PR #78, opened 2021-05-26
Add data channel transferable algorithm

youennf:
Fixes https://github.com/w3c/webrtc-extensions/issues/64

***
<a href="https://pr-preview.s3.amazonaws.com/youennf/webrtc-extensions/pull/78.html" title="Last updated on Jun 4, 2021, 9:29 AM UTC (40c0478)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-extensions/78/e706a8f...youennf:40c0478.html" title="Last updated on Jun 4, 2021, 9:29 AM UTC (40c0478)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -968,7 +968,6 @@ partial interface RTCDataChannel {
       <ol>
         <li><p>If <var>value</var>.`[[IsTransferable]]` is <code>false</code>, throw a "DataCloneError" DOMException.</p></li>
         <li><p>Set <var>dataHolder</var>.`[[ReadyState]]` to <var>value</var>.<a href="WEBRTC#dfn-readystate">`[[ReadyState]]`</a>.</p></li>
-        <li><p>Set <var>dataHolder</var>.`[[BufferedAmount]]` to <var>value</var>.<a href="WEBRTC#dfn-bufferedamount">`[[BufferedAmount]]`</a>.</p></li>
         <li><p>Set <var>dataHolder</var>.`[[DataChannelLabel]]` to <var>value</var>.<a href="WEBRTC#dfn-datachannellabel">`[[DataChannelLabel]]`</a>.</p></li>
         <li><p>Set <var>dataHolder</var>.`[[Ordered]]` to <var>value</var>.<a href="WEBRTC#dfn-ordered">`[[Ordered]]`</a>.</p></li>
         <li><p>Set <var>dataHolder</var>.`[[MaxPacketLifeTime]]` to <var>value</var>.<a href="WEBRTC#dfn-maxpacketlifetime"></a>.`[[MaxPacketLifeTime]]`</p></li>
@@ -984,7 +983,6 @@ partial interface RTCDataChannel {
     <div><p>The {{RTCDataChannel}} <a data-cite="!HTML/#transfer-receiving-steps">transfer-receiving steps</a>, given <var>dataHolder</var> and <var>channel</var>, are:</p>
       <ol>
         <li><p>Initialize <var>channel</var>.<a href="WEBRTC#dfn-readystate">`[[ReadyState]]`</a> to <var>dataHolder</var>.`[[ReadyState]]`.</p></li>
-        <li><p>Initialize <var>channel</var>.<a href="WEBRTC#dfn-bufferedamount">`[[BufferedAmount]]`</a> to <var>dataHolder</var>.`[[BufferedAmount]]`.</p></li>
         <li><p>Initialize <var>channel</var>.<a href="WEBRTC#dfn-datachannellabel">`[[DataChannelLabel]]`</a> to <var>dataHolder</var>.`[[DataChannelLabel]]`.</p></li>
         <li><p>Initialize <var>channel</var>.<a href="WEBRTC#dfn-ordered">`[[Ordered]]`</a> to <var>dataHolder</var>.`[[Ordered]]`.</p></li>
         <li><p>Initialize <var>channel</var>.<a href="WEBRTC#dfn-maxpacketlifetime">`[[MaxPacketLifeTime]]`</a> to <var>dataHolder</var>.`[[MaxPacketLifeTime]]`.</p></li>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -920,5 +920,69 @@ <h3>
...
+      This allows sending and receiving messages outside the context the connection was created, for instance in workers or third-party iframes.</p>
+    <div>
+      <p>The WebIDL changes are the following:
+      <pre class="idl"
+>[Exposed=(Window,Worker), Transferable]
+partial interface RTCDataChannel {
+};</pre>
+    </div>
+    <div>
+      <p>At creation of a {{RTCDataChannel}} object, called <var>channel</var>, run the following steps:</p>
```
jan-ivar, line 937: We should reference https://w3c.github.io/webrtc-pc/#dfn-create-an-rtcdatachannel explicitly here.

youennf:
https://w3c.github.io/webrtc-identity/#creating-an-offer could be used to do better monkey patching

index.html
youennf: Done

index.html
```diff
@@ -920,5 +920,69 @@ <h3>
...
+      <ol>
+        <li><p>Set <var>channel</var>.`[[IsTransferable]]` to <code>false</code>.</p></li>
+      </ol>
+    </div>
+    <div>
+      <p>The {{RTCDataChannel}} <a data-cite="!HTML/#transfer-steps">transfer steps</a>, given <var>value</var> and <var>dataHolder</var>, are:</p>
+      <ol>
+        <li><p>If <var>value</var>.`[[IsTransferable]]` is <code>false</code>, throw a "DataCloneError" DOMException.</p></li>
+        <li><p>Set <var>dataHolder</var>.`[[ReadyState]]` to <var>value</var>.<a href="WEBRTC#dfn-readystate">`[[ReadyState]]`</a>.</p></li>
+        <li><p>Set <var>dataHolder</var>.`[[BufferedAmount]]` to <var>value</var>.<a href="WEBRTC#dfn-bufferedamount">`[[BufferedAmount]]`</a>.</p></li>
```
alvestrand, line 957: I dont' think we should transfer bufferedAmount, because that value could change.

index.html
```diff
@@ -920,5 +920,69 @@ <h3>
...
+        <li><p>Set <var>dataHolder</var>.`[[DataChannelId]]` to <var>value</var>.<a href="WEBRTC#dfn-datachannelid">`[[DataChannelId]]`</a>.</p></li>
+        <li><p>Set <var>dataHolder</var>.`[[underlyingDataTransport]]` to <var>value</var> <a href="WEBRTC#dfn-data-transport">underlying data transport</a>.</p></li>
+        <li><p>Set <var>value</var>.`[[IsTransferable]]` to <code>false</code>.</p></li>
+        <li><p>Set <var>value</var>.`[[ReadyState]]` to <a data-cite="WEBRTC/#dom-rtcdatachannelstate-closed">"closed"</a>.</p></li>
+      </ol>
+    </div>
+    <div><p>The {{RTCDataChannel}} <a data-cite="!HTML/#transfer-receiving-steps">transfer-receiving steps</a>, given <var>dataHolder</var> and <var>channel</var>, are:</p>
+      <ol>
+        <li><p>Initialize <var>channel</var>.<a href="WEBRTC#dfn-readystate">`[[ReadyState]]`</a> to <var>dataHolder</var>.`[[ReadyState]]`.</p></li>
+        <li><p>Initialize <var>channel</var>.<a href="WEBRTC#dfn-bufferedamount">`[[BufferedAmount]]`</a> to <var>dataHolder</var>.`[[BufferedAmount]]`.</p></li>
```
alvestrand, line 973: I think this should instead say "Initialize BufferedAmount by querying the underlying data transport's state".

index.html
```diff
@@ -920,5 +920,69 @@ <h3>
               </dd>
             </dl>
 </section>
+<section>
+    <h3>
+      {{RTCDataChannel}} extensions
+    </h3>
+    <p>This document extends {{RTCDataChannel}} by making it <a data-cite="!HTML/#transferable-objects">transferable</a>.</p>
+      This allows sending and receiving messages outside the context the connection was created, for instance in workers or third-party iframes.</p>
```
alvestrand, line 928: I think this paragraph can be extended a little bit to give more context. In particular, something like:

An RTCDataChannel can be transferred immediately after being created. As soon as incoming data may be present on the RTCDataChannel, it is no longer transferable.

It is possible to call send() on a DataChannel before transferring it.

index.html
```diff
@@ -920,5 +920,69 @@ <h3>
...
+};</pre>
+    </div>
+    <div>
+      <p>The <a data-cite="WEBRTC#dfn-create-an-rtcdatachannel">create an RTCDataChannel</a> algorithm is updated by adding the following steps after step 4 of the original algorithm:</p>
+      <ol>
+        <li><p>Initialize <var>channel</var>.`[[IsTransferable]]` to <code>true</code>.</p></li>
+        <li><p>Queue a task to run the following step:</p>
+          <ol>
+            <li><p>Set <var>channel</var>.`[[IsTransferable]]` to <code>false</code>.</p></li>
+          </ol>
```
alvestrand, line 943: Suggest adding a paragraph saying "This task must be guaranteed to run before any incoming data is presented to "channel"." It is not obvious that the implementation has to enforce that property; in theory there could be ondata events queued behind the onopen event before the onopen event gets started; this transfer algorithm has to ban that implementation strategy.

alvestrand (commented):
Looks good, have some comments I think we should address to make requirements clearer.

index.html
youennf: That value should always be 0, otherwise channel is not transferable

index.html
youennf: For simplicity, calling send() will not make the data channel transferable anymore.

index.html
alvestrand: Then there's no need to transfer it.

Good that you added the prohibition against send() explicitly - I was reviewing the version before that change.

index.html
youennf: Sounds fair, I'll add something

index.html
alvestrand: Or even simpler: "set it to zero".

index.html
alvestrand: Or even simpler: Don't include it in the dataHolder.

index.html
aboba: Agree that allowing calling `send()` should probably be disallowed for simplicity. I'm more worried about data that may be in the receive pipeline.  Data can be sent by the other peer as soon as the SctpTransport is established and the data channel is open. Why force the queued data to be thrown away?

youennf: As discussed during the call, we make sure no message data is lost in the transfer.
This is explicitly stated in the latest PR.

index.html
youennf: Removed the line, according yesterday's discussion

index.html
youennf: Done

Issue #73, opened 2021-04-22
Share some connection infrastructure with Fetch


annevk:
I think there are basically two reasons why we would want centralized connection management in the platform:

* Consistent handling of overrides of resolving names (domains), e.g., mapping localhost to the loopback address, as discussed at https://github.com/whatwg/fetch/issues/1118.
* Consistent requirements on partitioning of all aspects of the connection. https://fetch.spec.whatwg.org/#network-partition-keys in Fetch is now used for connections and the HTTP cache, but it should also apply to the DNS cache, etc. Having those kind of requirements once would be beneficial. Also in case we ever decide to adjust them.

(See also https://github.com/w3c/webtransport/issues/128.)

alvestrand:
The network partition key (https://fetch.spec.whatwg.org/#network-partition-key) may be a sharable object.

jan-ivar:
@annevk to unstuck this, would you be able to provide examples or guidance (or even a PR) with how and which algorithms you imagine WebRTC invoking in fetch here?

annevk:
I'm slowly making progress on enabling this together with @ricea. The idea is that WebRTC would at some point hook into https://fetch.spec.whatwg.org/#concept-connection-obtain (or perhaps something lower-level as mt points out above), with parameters that make sense for it. That would then take care of any domain resolving (if needed) and setting up and returning a connection.

My hope is that we can also move some other logic there, such as port blocking, what `localhost` means, etc. (I think since mt's comment above implementations have reached agreement on sharing port blocking logic across HTTP/WebRTC/WebTransport connections, with the exception of one or two ports that are safelisted for WebRTC use.)

alvestrand:
Anne, I'm having a bit of trouble understanding how "obtain a connection" could be a common merge point.
To start off with, WebRTC doesn't use hostnames in the common use case (it does use MDNS hostnames in some cases, and it's possible to use hostnames too, but the by-far majority of usages use IP addresses).

I do appreciate that @ricea is trying to help, but having someone who knows the WebRTC networking stack (ICE, STUN and all that machinery) in the loop is probably vital, so that we don't generate proposals that don't make sense in the WebRTC context.

annevk:
Agreed. We're still in the process of ensuring Fetch, WebSocket, and WebTransport are aligned and it might make sense that WebRTC only reuses some aspects. I'd love to chat once we're closer. Here are the things that come to mind with regards to sharing:

[...]

PR #81, opened 2021-05-30
Add a CSP check to RTCPeerConnection.constructor().

zenhack:
Corresponding change to CSP spec:

https://github.com/w3c/webappsec-csp/pull/457

I'm like 90% sure I've botched the formatting here in one way or another; I still haven't totally gotten my head around how links are generated in these docs, and it seems like there's a bit of inconsistency in how docs are build from spec to spec... so let me know if something's broken...

I'm also a little fuzzy on how the CSP doc should refer to this; right now it links to the constructor section in WebRTC 1.0, but of course this step isn't there, so maybe it should link to the extensions instead?

cc: @annevk, @jan-ivar

### DIFF
--- a/index.html
+++ b/index.html
@@ -101,7 +101,8 @@
     <p>
       The {{RTCPeerConnection}}
       interface is defined in [[WEBRTC]]. This document extends that interface
-      by adding an additional static method.
+      by adding an additional static method, and adding a Content-Security-Policy
+      check to the constructor.
     </p>
     <pre class="idl">partial interface RTCPeerConnection {
   static sequence&lt;RTCIceServer&gt; getDefaultIceServers();
@@ -138,6 +139,13 @@
           </div>
         </dd>
       </dl>
+    </section>
+    <section>
+      <h2>Modifications to existing procedures</h2>
+
+      <p>In the <a data-cite="WEBRTC#dom-peerconnection">Constructor</a> algorithm,
+      add a step right before step 2 saying "Run [[WEBAPPSEC-CSP#should-block-rtc-connection]]
+      on the current global object. Throw a {{NotAllowedError}} if it returns `Blocked`."</p>
     </section>
   </section>
   <section id="rtcrtptransceiver-interface">


annevk:
I would prefer we add this to https://github.com/w3c/webrtc-pc directly. Monkey patching standards tends to lead to problems.

zenhack:
I agree that would be cleaner. I'd gotten the impression somewhere that that document was more set in stone, but if you think it's ok I'll find time to put together a pr in the next couple days.

jan-ivar:
> > I would prefer we add this to https://github.com/w3c/webrtc-pc directly. Monkey patching standards tends to lead to problems.
> 
> ... I'd gotten the impression somewhere that that document was more set in stone, ...

@annevk @zenhack Yes, WebRTC is at _Recommendation_, and this WG is still finding its legs wrt a living spec within the W3C. 

This PR is a trial-balloon for our new [merge guide](https://github.com/w3c/webrtc-pc/blob/main/merge-guide.md), and I would say it's not going well, as it looks trivially small, yet missed the cutoff we made for small bug-fixes vs features.

index.html
```diff
@@ -139,6 +140,13 @@ <h2>Methods</h2>
         </dd>
       </dl>
     </section>
+    <section>
+      <h2>Modifications to existing procedures</h2>
+
+      <p>In the <a data-cite="WEBRTC#dom-peerconnection">Constructor</a> algorithm,
+      add a step right before step 2 saying "Run [[WEBAPPSEC-CSP#should-block-rtc-connection]]
+      on the current global object. Throw a {{NotAllowedError}} if it returns `Blocked`."</p>
```
jan-ivar, line 148: To avoid referencing step numbers, we've been recommending the [following pattern](https://w3c.github.io/webrtc-identity/#creating-an-offer) (it might limit us to adding to the top and bottom of algorithms, but we'll see how it goes):

```suggestion
      <p>Whenever the <a data-cite="WEBRTC#dom-peerconnection">RTCPeerConnection constructor</a>
      algorithm is invoked, run the following steps instead:
        <ol>
          <li>If [[WEBAPPSEC-CSP#should-block-rtc-connection]] on the current global object returns "Blocked",
          throw a {{NotAllowedError}}."</li>
          <li>Return the result of running the original
          <a data-cite="WEBRTC#dom-peerconnection">RTCPeerConnection constructor</a> algorithm.</li>
        </ol>
      </p>
```

Also, using `[[` `]]` appears to not work well for some reason, or is that a limitation of Preview?

jan-ivar (changes requested):

jan-ivar:
> ... yet missed the cutoff we made for small bug-fixes vs features. ...

Unless we can get it in under _"patch significant security holes in WebRTC"_?

index.html
zenhack: It works well enough in this case, done. Re: `[[` `]]`, it could be that the name of the doc is wrong (it was a shot in the dark, I'm not sure how to locate the correct name). Or it could be something else; as I said in the original post I don't totally grok how all of the formatting for these docs works.

zenhack:
I have no objection to using the security holes clause to justify merging this directly into `webrtc-pc`; if folks can come to a consensus I'll follow.

jan-ivar:
@alvestrand @aboba WDYT?

index.html
```diff
@@ -139,6 +140,19 @@ <h2>Methods</h2>
...
       </dl>
     </section>
+    <section>
+      <h2>Modifications to existing procedures</h2>
+
+      <p>Whenever the <a data-cite="WEBRTC#dom-peerconnection">RTCPeerConnection constructor</a>
+      algorithm is invoked, run the following steps instead:
+        <ol>
+          <li>If [[WEBAPPSEC-CSP#should-block-rtc-connection]] on the current global object returns "Blocked",
+          throw a {{NotAllowedError}}."</li>
```
jan-ivar, line 150: Should this be `SecurityError` to match the 'S' in CSP?

alvestrand:
I'm afraid I have an objection to using the "security hole" merge-fix loophole.
This is new functionality, it is not implemented in any browser, and has not been anyone's high priority since I last messed with it in 2018 (when it died for lack of compelling interest).

Our agreement when we did the merge-guide was that we'd keep the idea that REC documented stuff for which there was consensus agreement and had been implemented in at least 2 browsers. Anything else goes to -extensions.

index.html
```diff
@@ -139,6 +140,19 @@ <h2>Methods</h2>
...
+          <li>Return the result of running the original
+          <a data-cite="WEBRTC#dom-peerconnection">RTCPeerConnection constructor</a> algorithm.</li>
```
alvestrand, line 152: The constructor algorithm has its own tag: https://w3c.github.io/webrtc-pc/#constructor

It's probably to use that rather than linking to the whole peerconnection definition.

index.html
```diff
@@ -139,6 +140,19 @@ <h2>Methods</h2>
         </dd>
       </dl>
     </section>
+    <section>
+      <h2>Modifications to existing procedures</h2>
+
+      <p>Whenever the <a data-cite="WEBRTC#dom-peerconnection">RTCPeerConnection constructor</a>
+      algorithm is invoked, run the following steps instead:
+        <ol>
+          <li>If [[WEBAPPSEC-CSP#should-block-rtc-connection]] on the current global object returns "Blocked",
```
alvestrand, line 149: Query: the constructor steps have a [[DocumentOrigin]] initialized to the relevant settings object's origin. Is "the relevant settings object" the same as "the current global object", or are they different? And if they're different, is this intentional?

index.html
```diff
@@ -101,7 +101,8 @@ <h3>
     <p>
       The {{RTCPeerConnection}}
       interface is defined in [[WEBRTC]]. This document extends that interface
-      by adding an additional static method.
+      by adding an additional static method, and adding a Content-Security-Policy
```
alvestrand, line 104: I don't see the additional static method in this PR. Did it get deleted again?

alvestrand (commented):
Still not 100% convinced that this is a good idea. But let's get it properly documented.

index.html
annevk: In the context of a constructor **this's relevant ...** and **current ...** are identical, though a settings object is different from a global object. https://html.spec.whatwg.org/multipage/webappapis.html#realms-settings-objects-global-objects has a longer explanation on this. Sticking with one pattern is better though, I'd recommend sticking with **this's relevant ...**.

aboba:
@alvestrand I agree that we cannot add this directly to webrtc-pc due to lack of implementation. So it should live here (with a "feature at risk" note) until it is implemented.

index.html
zenhack: This is referring to getDefaultIceServers(), which is added by -extensions, but not by this pr -- only the  ", and adding a Content-Security-Policy" part of this line is new.

index.html
zenhack: should-block-rtc-connection says it takes a global object, and it ends up passing that through to create-violation-for-global, so it seems like passing in a settings object is conceptually a type error here, and changing it on the other end would be fairly invasive. But I'm still a little fuzzy on how all these relate, so it's possible I'm missing something.

zenhack:
Addressed/replied to all outstanding comments.

I believe I still need to sign the ipr commitment for this repo -- I think for webappsec-csp @samuelweiler sent me the relevant form, not sure if someone needs to intervene to send me one for this repo or if it's up somewhere public that I haven't found.

annevk:
So this makes the constructor throw, which is a bit different from how CSP works with other network APIs. There CSP violations effectively ends up causing a network error. Was that considered here? That might also make this more compatible in the couple cases where it might cause issues in the sense that applications would not end up getting an exception they might not have accounted for, but would rather get network errors they likely deal with already anyway.

zenhack:
My rationale was that putting it here makes everything simpler -- there's one entry point, so this just cuts the whole thing off at the source, which is easier to reason about than if we had to track down all of the places where network activity might happen lazily. Though this opens up the question of what should we do if we find ourselves wanting to add finer-grained permissions?

This is  perhaps another argument in favor of using Permissions-Policy or such instead of CSP, though that doesn't solve the practical problem of errors where apps don't expect them.

zenhack:
Outstanding issues as I see them:

- @alvestrand says he's not sold on the idea. Could you elaborate on your objection?
- @annevk raised concerns regarding new places exceptions may be raised. Relevant things to decide there:
  - Are we ok with a new exception being thrown if an operator sets this?
  - Do we think pushing the error reporting later is worth the extra complexity, and if so where would this go instead?
  - If we keep the current design, how might this evolve if we make the csp directive finer-grained?

I'm of the opinion that the additional complexity is *not* worth it for the error handling consideration alone, but am interested in hearing other perspectives. I don't have a great answer to the evolvability question though. I also am not sure where the check should go if we push it later -- if others have insights please share. My biggest concern with pushing the check later is I want to have a clear story as to how we will have confidence that we haven't missed a spot.

zenhack:
Ping?

alvestrand:
Not sold on the idea:

- I'm not sure CSP is the right interface to modify in order to restrict WebRTC usage; my conversations with @mkwst back in 2018 indicated that there was some thought that CSP was already too complex and that better solutions were coming.
- I saw this idea brought up in 2018 and die for lack of a customer that really wanted it; I'm not sure we have a strong enough customer to get it implemented across browsers this round.

That said, I'm fine with the proposal technically - there are a number of usages of WebRTC that don't involve communication (attempts at inappropriate user tracking, mostly), and I'm not at all unhappy about blocking them rather than waiting until we can signal a "network failure" as a response to an attempt at communication.

So I won't block this at all, but I'm somewhat skeptical of it getting through the process and getting deployed.

zenhack:
> I saw this idea brought up in 2018 and die for lack of a customer that really wanted it; I'm not sure we have a strong enough customer to get it implemented across browsers this round.

This is pretty much mandatory for where Sandstorm wants to go; if I have to I'll send a patch to all three major FOSS browser engines myself, though it would obviously more efficient if someone already familiar with the relevant codebases had the bandwidth.

alvestrand:
What's Sandstorm, and where does it want to go?

zenhack:
See https://github.com/w3c/webrtc-extensions/issues/64#issuecomment-764922669 (which was in response to a separate but related issue for which there seems to be a broader interest)

annevk:
With regards to pushing the check later: #73 is about sharing more infrastructure with Fetch. once we share connection setup infrastructure across all (web platform) specifications, we also have a single point to block them. It would just act as if establishing the connection resulted in failure.

rumkin:
From the perspective of Sandstorm and reliable sandboxing, there should be 'service-worker-src' directive to allow workers and shared workers to be provided by third party and service worker provided only by trusted source. This is because of ability to service worker to override security headers and thus to escape from the sandbox.

zenhack:
> This is because of ability to service worker to override security headers and thus to escape from the sandbox.

Can you expand on this? I'm not deeply familiar with service workers, how might they do this?

There does appear to be a `worker-src` directive: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/worker-src

rumkin:
Service worker allows you to intercept requests to server and reply with a newly created Response from the browser itself (see [FetchEvent](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent)). New response object is instance of [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) and thus can contain CSP header. What allows it to override current CSP policy. So having third-party service worker is not safe even with restrictive CSP.

zenhack:
Thanks for the heads up. Probably further discussion of `service-worker-csp` should get its own issue.

zenhack:
So one other place I think I see to enforce this is the notion of administratively prohibited ice candidates described in the 1.0 spec. This has the advantage over the constructor that its somewhere the spec already allows failure, and it makes it clear how we would extend it to filtering ice candidates, turn servers, etc. Two questions:

1. Am I correct in thinkikng that simply treating all candidates as administratively prohibited would effectively stop webrtc from doing anything with the network?
2. Do others think this is a good aproach?

lgrahl:
> So one other place I think I see to enforce this is the notion of administratively prohibited ice candidates described in the 1.0 spec. This has the advantage over the constructor that its somewhere the spec already allows failure, and it makes it clear how we would extend it to filtering ice candidates, turn servers, etc. Two questions:
> 
>     1. Am I correct in thinkikng that simply treating all candidates as administratively prohibited would effectively stop webrtc from doing anything with the network?
> 
>     2. Do others think this is a good aproach?

A word of warning when it comes to filtering: As mentioned way back in https://github.com/w3c/webappsec-csp/pull/287#issuecomment-366446690, more than an on/off switch is more likely to provide a false sense of security than anything else.

zenhack:
@annevk, @alvestrand, does my suggestion above seem reasonable to you?

annevk:
Yeah I think that works and I favor that approach as it gives callers a single failure signal.

alvestrand:
Treating all remote candidates as adminstratively prohibited (ie not causing a connection attempt) would indeed prevent RTCPeerConnection from making a connection to a remote site. As outlined in https://w3c.github.io/webrtc-pc/#dfn-administratively-prohibited, it would not give any signal to the user on why the connection failed.

It does not prevent usage of an RTCPeerConnection for strictly local purposes, like trying to gather local interface information, but since that's not the goal of CSP checks, I think that's OK.

zenhack:
Ok, sounds like a plan. It's not immediately obvious how to go about adding such a change to the document, since everything else in the extensions doc either touches specific interfaces or is non-normative. Should we just add a section under overviews? Or perhaps it would be better to add a new major section for definitional changes like this? I'm interested in others' thoughts on this, open to suggestions.

(I kinda wish the extensions doc was a set of actual (machine-readable) patches to the main doc; it seems like it would be much easier to maintain. But that's obviously very out of scope for this pr).

alvestrand:
(The chairs are discussing possible means of doing inline patches rather than extensions docs - going IEEE style rather than IETF, but hopefully better supported. The main worry is that we'd be breaking uncharted ground, and introducing Yet Another Special Markup Syntax. But that's another issue.)

The traditional "monkey patch" approach is to say something like "between steps A and B in algorithm X, insert the following steps..."

zenhack:
Ok, I finaly got around to doing another pass, with the discussed change. PTAL.

(I also changed the title of the issue accordingly).

index.html
```diff
@@ -139,6 +140,17 @@ <h2>Methods</h2>
...
       </dl>
     </section>
+    <section>
+      <h2>Modifications to existing procedures</h2>
+
+      <p>Append the following paragraph to the description of
+      <a data-cite="WEBRTC#dom-peerconnection-addicecandidate">addIceCandidate</a>:</p>
+
+      <p>If <a data-cite="WEBAPPSEC-CSP#should-block-rtc-connection">
+        should-block-rtc-connection</a> on the current global object returns
```
annevk, line 150: This link doesn't seem to be working. At least not when I hit the Preview link in OP.

annevk (commented):
This model seems reasonable to me.

index.html
zenhack: I assume that's because its target is introduced in https://github.com/w3c/webappsec-csp/pull/457, which hasn't been merged yet.

alvestrand:
Hm. With the block in addIceCandidate, that means that we can still talk to TURN and STUN servers (ask them for candidates).
Should that possibility be closed down too?

lgrahl:
To be honest, I find this approach kind of a hack and believe it's desirable to just prohibit creating `RTCPeerConnection` instances:

1. It's confusing that one can create a peer connection at all if WebRTC was _disabled_ (I know the policy is described differently now) in the CSP policy. I mean, what's the point? Or is there a use case I'm missing?
2. Harald makes a good example on how easy it is to miss something that still allows to ping somewhere via WebRTC. I'm imagining there are further loopholes if one is just crafty enough, potentially via the use of SDP, e.g. spoofing local mDNS ICE candidates.

I'd rather not enter that rabbit hole and just fence it off entirely, at least with the on/off switch that it's supposed to be.

zenhack:
@lgrahl, I still myself have a preference for just throwing in the constructor, but @annevk objected to introducing a new place where errors could occur. I agree it would be much simpler to get the language right in that case.

@annevk, how strongly do you feel about this? I'm doubtful that there are many real world use cases where the extra error will be a problem; it would be limited to cases where an app tries to use WebRTC when the server operator has disabled it, which is already likely to be rare, and I can't imagine such an app could do more than fail with a better error if the authors have an existing error handler. I'm willing to defer to consensus here, but this very much feels like the wrong side of the trade-off to me.

---

wrt @alvestrand's comments, I agree we should tweak the language to more explicitly block all network communication. I believe with this interface an implementation shouldn't ever *need* to do network communication; it should be able to just skip even looking for ICE candidates, since it knows all of them will be filtered out. But we should spell out that that behavior is required.

I'll see if I can fuss with this soonish. That said, @annevk if you are at all open to the simpler solution of throwing in the constructor I still like that better.

alvestrand:
There are a number of cases that will throw in the constructor for various reasons already. https://w3c.github.io/webrtc-pc/#constructor lists several, including the catch-all "if anything else fails, throw an UnknownError".

zenhack:
Hm, yes, that's true. I'm surprised in hindsight that I didn't push on
this at the time...

Quoting Harald Alvestrand (2021-11-18 17:45:13)

annevk:
The model I aimed for is consistency with the other networking APIs. All failures there are essentially surfaced as if there was a network error. This could be a DNS error, a CORS failure, a CSP failure, blocking of Mixed Content, etc.

It seemed to me that if we put this check in the same place that we block ports and such, it should be okay, but apparently there's other ways? Do those other ways end up bypassing other checks they shouldn't?

jan-ivar:
> The model I aimed for is consistency with the other networking APIs. All failures there are essentially surfaced as if there was a network error. This could be a DNS error, a CORS failure, a CSP failure, blocking of Mixed Content, etc.

@annevk What are the underlying requirements met by this model?

RTCPeerConnection differs from more linear connection establishment APIs in several ways:
1. It talks to [up to 64](https://github.com/w3c/webrtc-pc/pull/2679) STUN/TURN servers provided by JS, with no error path anyone listens to ([onicecandidateerror](https://wpt.fyi/results/webrtc/RTCPeerConnection-onicecandidateerror.https.html)). So lots of networking activity normally happens (STUN/TURN requests + out-of-band-signaling in JS) on both sides, before we ever reach ["failed"](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnectionstate-failed), which may be seconds, [if at all](https://github.com/w3c/webrtc-pc/issues/2698).
2. ["failed"](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnectionstate-failed) isn't a terminal state, since JS can [pc.restartIce()](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-restartice) ([diagram](https://w3c.github.io/webrtc-pc/#fig-non-normative-ice-transport-state-transition-diagram)).
3. ["closed"](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnectionstate-closed), the only terminal state, is only reachable from [pc.close()](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-close) (never surfaced to JS e.g. through an event).

In other words, RTCPeerConnection is more of a p2p connection negotiation factory than a connection itself, and doesn't really die until you kill it.

We can still try to find a common network failure path to simulate, like halting the ICE agent entirely — though when I tried, I fell in mud, and had to file https://github.com/w3c/webrtc-pc/issues/2698  — we can fix that, but it'd be good to know if it meets the requirements.

annevk:
@jan-ivar I'm not sure if they are requirements, but it has some nice properties:

* By having a single error signal, we reduce the risk of exposing privileged information.
* By having a single error signal, we simplify the setup for web developers and reduce the chance an exceptional situation isn't handled by code.

When it talks to STUN/TURN servers, isn't at least port blocking in effect there? It seems weird if there are no policies governing those connections.

lgrahl:
> When it talks to STUN/TURN servers, isn't at least port blocking in effect there? It seems weird if there are no policies governing those connections.

From the [spec](https://w3c.github.io/webrtc-pc/#dom-peerconnection-addicecandidate):

>  A candidate is administratively prohibited if the UA has decided not to allow connection attempts to this address.
> 
> For privacy reasons, there is no indication to the developer about whether or not an address/port is blocked; it behaves exactly as if there was no response from the address.
> 
> The UA MUST prohibit connections to addresses on the [Fetch] block bad port list, and MAY choose to prohibit connections to other addresses. 

That being said, IIRC one can still exfiltrate data to arbitrary servers by crafting TURN credentials if that is left open. Furthermore, I'm almost certain that, due to the complexity of the state machine, implementations would have to place a whole bunch of "WebRTC is turned off" guard checks, so one can interact with the PC state machine safely in the "off" mode.

annevk:
> That being said, IIRC one can still exfiltrate data to arbitrary servers by crafting TURN credentials if that is left open.

Why wouldn't port restrictions apply here?

> Furthermore, I'm almost certain that, due to the complexity of the state machine, implementations would have to place a whole bunch of "WebRTC is turned off" guard checks, so one can interact with the PC state machine safely in the "off" mode.

Can you expand on this? Why would this be different from PC not being able to establish any connection due to them all using blocked ports?

lgrahl:
Maybe we have a different understanding of "blocked ports". Do you mean some (in this case UDP/TCP) ports are blocked or does *port* mean something else?

annevk:
UDP/TCP ports being blocked. And in particular this happening before there is network activity. (Defined in Fetch, modulo a number of issues.)

lgrahl:
Expanding on what I mean with guard checks then: If the port restriction applies to all connections for the whole port range, then there should be no problem on spec level. Of course, that's different in practical implementations since there isn't just _the one way_ to create sockets and initiate connections. For example, last time I checked Firefox used an ICE library underneath. Hacking this into the library could not only be painful but it's also easy to miss something. It's likely easier to fence off PC methods entirely and simulate a network failure - that means guard checks. And at that point, we need to ask if it's worth the effort.

But beyond that, as a developer I would also rather like to have explicit signals that WebRTC is turned off. Finding it out by listening for an ICE failure (which can take many seconds as @jan-ivar mentioned) isn't great.

annevk:
The port restrictions apply to all connections, yes. Not all ports are blocked though. Are you suggesting there are code paths specifically for non-blocked ports? That sounds rather wonky.

It's not clear to me we want to reveal the CSP restrictions directly to a potential attacker. If the website wanted to convey its CSP policy it can already, right?

lgrahl:
> Are you suggesting there are code paths specifically for non-blocked ports? That sounds rather wonky.

It could very well be that I'm wrong on that. What I did was look at it from a top-down perspective. There's just a lot more places in WebRTC where connections are spawned than for example in the usual HTTP client implementation. Maybe Harald or Jan-Ivar can clarify better if that would be hard to implement.

> It's not clear to me we want to reveal the CSP restrictions directly to a potential attacker. If the website wanted to convey its CSP policy it can already, right?

Ok, fair.

alvestrand:
Pardon my ignorance about CSP mechanics...

my impression was that CSP policies were conveyed in HTTP headers, and that HTTP headers are as a general rule visible to the page that is created from the HTML that is carried under those HTTP headers.

Is there a rule that CSP headers are rendered invisible to the page by the browser?

If not, we can just assume that the CSP policy is visible to the page, and making its effects easily visible to the user is not revealing additional information to a competent attacker.
If yes, we need to take care not to reveal the restrictions.

annevk:
@alvestrand generally script cannot observe the headers a document is served with. (It could fetch the document, if it has that capability, but that might also yield a different result these days as `Sec-Fetch-Mode` would not be "`navigate'".)

zenhack:
So, from my perspective, I am more concerned by the additional complexity introduced by trying to do the filtering later in the pipeline than I am exposing the CSP policy to client javascript.

For the former, the failure mode is clear: we miss a place where something needs to be blocked, and so it is possible to exfiltrate data. The negative consequences of leaking this bit of CSP to the client are much less obiviously (to me) terrible, and my instinct is that it isn't worth the extra complexity to make sure the app can't tell the difference between webrtc being disabled and the network being flaky.

To try to keep this moving forward:

@annevk,

1. How strongly do you feel about attempting to emulate failure modes for other network APIs?
2. Are there concrete problems with exposing this bit of CSP, or is it just the general principle of "better to expose less?"

@alvestrand/@jan-ivar/other webrtc people, who are more familiar with the implementations:

* Are there places other than the RTCPeerConnection constructor that seem like good candidates to be the choke point, i.e. places where we could add a single local check to block everything?
* I think the addIceCandidates works from an *API* perspective, but what I'm inferring is that an implementation that avoids doing anything with the network would probably be more complex in this case. Have I understood that correctly, or is this still something that would be relatively simple to implement?

annevk:
I feel pretty strongly about it, especially as we already need to be able to enforce policies on those connections, so this shouldn't increase complexity. Rationale is in https://github.com/w3c/webrtc-extensions/pull/81#issuecomment-973853662 and a subsequent comment.

zenhack:
OK - I'll keep an eye on webrtc-pc#2698; if @jan-ivar's idea of where to have a single cutoff works I'm happy.

zenhack:
Alright, as I understand it, now that @jan-ivar's pr is merged, we are unblocked on this. I guess what is needed now is some language somewhere to the effect that implementations should not trigger any traffic if CSP is blocking webrtc, but I'm a bit fuzzy on what exactly that should look like. I may find time late in the week to stare at the problem more, but if anyone has a clear idea of what this could look like I am interested in hearing it.

zenhack:
I'm really scratching my head re: where exactly add to specify the the "no network traffic" we want here. What seems clearest to me is to add an overview section noting this, but iiuc overview sections are non-normative, and I'm not sure where to put something like this that isn't attached to the behavior of a specific interface, but rather a prescription of global behavior. What is the right way to do this kind of thing?

zenhack:
Sounds like a good idea; I did a pass at it, PTAL.

I hadn't really processed that the definition of administratively prohibited was local to addIceCandidate; that makes it easier to work into the extensions document.

zenhack:
Ack, now that I've actually looked at your pr I understand that your intention was to just fix the issue there; I guess I should back out the change I just made?

index.html
```diff
@@ -139,6 +142,21 @@ <h2>Methods</h2>
         </dd>
       </dl>
     </section>
+    <section>
+      <h2>Modifications to existing procedures</h2>
+
+      <p>Append the following two paragraphs to the description of
+      <a data-cite="WEBRTC#dom-peerconnection-addicecandidate">addIceCandidate</a>:</p>
```
jan-ivar, line 149: We might move the algorithm to a neutral chapter in a separate editorial PR, so I'd remove mention of addIceCandidate here.
```suggestion
      <p>Append the following paragraph to the
      <a data-cite="WEBRTC#dfn-administratively-prohibited">administratively prohibited</a>
      algorithm:</p>
```

index.html
```diff
@@ -139,6 +142,21 @@ <h2>Methods</h2>
...
+
+      <p>If <a data-cite="WEBAPPSEC-CSP#should-block-rtc-connection">
+        should-block-rtc-connection</a> on the current global object returns
+      "Blocked", all candidates MUST be <a data-cite="WEBRTC#dfn-administratively-prohibited">
+        administratively prohibited</a>.</p>
+
+      <p>If a candidate is <a data-cite="WEBRTC#dfn-administratively-prohibited">,
+        the UA MUST NOT provide the candidate to the
+        <a data-cite="WEBRTC#dfn-ice-agent">ICE Agent</a>.</p>
```
jan-ivar, line 158: Once https://github.com/w3c/webrtc-pc/pull/2708 is merged (assuming it is merged as currently written), we can remove this, but I suppose it doesn't hurt to leave it here until then.

jan-ivar (approved):
> I hadn't really processed that the definition of administratively prohibited was local to addIceCandidate

That's a mistake in the spec actually, and is issue https://github.com/w3c/webrtc-pc/issues/2675 for which I've provided the PR. With the PR the algorithm is invoked from *addIceCandidate*, *setRemoteDescription* and *setLocalDescription*. The first two matches the reality that web developers who know SDP may provide candidates to the browser through both methods, and SLD is what triggers local gathering.

index.html
zenhack: Since that pr is merged, I've backed this out.

index.html
zenhack: Done.

jan-ivar (approved):

jan-ivar:
OK to merge once https://github.com/w3c/webappsec-csp/pull/457 is merged.

zenhack:
Fixed the merge conflicts, and https://github.com/w3c/webappsec-csp/pull/457 has been merged.

zenhack:
Anything still need to happen here?

zenhack:
@jan-ivar, ping?

zenhack:
I think somebody just needs to push the button...

index.html
```diff
@@ -95,6 +95,28 @@ <h3>RTP header control</h3>
...
+    <section>
+      <h2>Modifications to existing procedures</h2>
+
+      <p>Append the following paragraph to the
+      <a data-cite="WEBRTC#dfn-administratively-prohibited">administratively prohibited</a>
+      algorithm:</p>
+
+      <p>If <a data-cite="WEBAPPSEC-CSP#should-block-rtc-connection">
+        should-block-rtc-connection</a> on the current global object returns
+      "Blocked", all candidates MUST be <a data-cite="WEBRTC#dfn-administratively-prohibited">
```
jan-ivar, line 116: Thanks @zenhack for your patience! — I've modified the cite to get it to link (should work once TR updates), and I've specified the required argument. If this LGTY I'll merge it.
```suggestion
      <p>If <a data-cite="CSP#should-block-rtc-connection">
        should RTC connections be blocked for global?</a> with the
      [=relevant global object=] of the {{RTCPeerConnection}} object in question
      returns `"Blocked"`, then all candidates MUST be <a
        data-cite="WEBRTC#dfn-administratively-prohibited">
```

zenhack:
Merged the change. The link looks right now, though the doc it's pointing to seems to be a version of the doc that doesn't have that section updated. LGTM though.

jan-ivar:
Checks seem to be failing over `error: Duplicate ID “rtcpeerconnection-interface”.` which seems unrelated to this PR (reuse is in "Removed features" section added in https://github.com/w3c/webrtc-extensions/pull/105). I'll see if I can patch it here.

Issue #130, opened 2022-12-02
Potential web compat issue with setOfferedRtpHeaderExtensions


jan-ivar:
I'm concerned about web compat over [this line](https://w3c.github.io/webrtc-extensions/#dom-rtcrtptransceiver-setofferedrtpheaderextensions):
- _"2. If ext.[uri](https://www.w3.org/TR/webrtc/#dom-rtcrtpheaderextensioncapability-uri) value is not present in [[[HeaderExtensionsToOffer]]](https://w3c.github.io/webrtc-extensions/#dfn-headerextensionstooffer), [throw](https://webidl.spec.whatwg.org/#dfn-throw) a [NotSupportedError](https://webidl.spec.whatwg.org/#notsupportederror)."_

...where [[HeaderExtensionsToOffer]] is [implementation-defined](https://infra.spec.whatwg.org/#implementation-defined). If _uri_ is a fixed string, that's a potential web compat headache.

How is JS expected to use this API in a web compatible manner?
 1. Come up with a subset tested to work in all browsers (today)?
 2. Detect UA and pick from a map?
 3. Write JS to pre-filter their input to [setOfferedRtpHeaderExtensions](https://w3c.github.io/webrtc-extensions/#dom-rtcrtptransceiver-setofferedrtpheaderextensions) using [headerExtensionsToOffer](https://w3c.github.io/webrtc-extensions/#dom-rtcrtptransceiver-headerextensionstooffer)?

I found no example showing predicted use of this API, but I could imagine some sites being lazy and setting a fixed array and not testing it in all browsers before going to production. This might cause calls to outright fail in a particular browser; or even work today but fail later once some browser makes a change.

This might also create a situation where browsers find it difficult to make changes later.

Throwing also seems redundant, since JS can already read [[HeaderExtensionsToOffer]] using [headerExtensionsToOffer](https://w3c.github.io/webrtc-extensions/#dom-rtcrtptransceiver-headerextensionstooffer), and can query [headerExtensionsNegotiated](https://w3c.github.io/webrtc-extensions/#dom-rtcrtptransceiver-headerextensionsnegotiated) for the net result.

Given this, maybe it's better for web compat to allow user agents to ignore values not present in [[HeaderExtensionsToOffer]]?

alvestrand:
To figure out which header extensions are available:
Read the HeaderExtensionsToOffer attribute, it will contain all header extensions that are supported.

[...]

Issue #137, opened 2022-12-09
FrozenArray on headerExtensionsToOffer a mistake, wrong semantics


jan-ivar:
With no explainer or examples showing how `headerExtensionsToOffer` is meant to be used, it's a bit unclear what its requirements are. But https://github.com/w3c/webrtc-extensions/issues/130#issuecomment-1338005567 suggests at least two possible uses I arrived at without any guidance:
1. to filter input meant for setHeaderExtensionsToOffer
2. to serve as a (modifiable) base for input to setHeaderExtensionsToOffer

However, once https://github.com/w3c/webrtc-extensions/issues/136 is fixed, using the output as input is going to no-op in ways that might surprise users:
```js
const headers = transceiver.headerExtensionsToOffer; // not using structuredClone() as workaround!
for (const header of headers) {
  if (header.uri == "urn:ietf:params:rtp-hdrext:ssrc-audio-level") {
    header.direction = "stopped";  // won't work if direction is unmodifiable
  }
}
transceiver.setOfferedRtpHeaderExtensions(headers);
```
JS is going to try to modify headers that are unmodifiable, not get any errors, and call setOfferedRtpHeaderExtensions thinking everything is fine, when nothing got done.

### Proposals
- The simplest fix seems to be to replace it with a familiar `getHeaderExtensionsToOffer()` returning a sequence\<dictionary>

henbos:
Completed as part of https://github.com/w3c/webrtc-extensions/pull/142

Issue #136, opened 2022-12-09
headerExtensionsToOffer entries are mutable which is surprising


jan-ivar:
The following behavior seems unintentional and surprising:
```js
if (transceiver.headerExtensionsToOffer[0].direction == "sendrecv") {
  transceiver.headerExtensionsToOffer[0].direction = "stopped";
  console.log(transceiver.headerExtensionsToOffer[0].direction); // "stopped"
}
```
This behavior is from the [spec](https://w3c.github.io/webrtc-extensions/#rtcrtptransceiver-interface) using FrozenArray on a dictionary:
```webidl
readonly attribute FrozenArray<RTCRtpHeaderExtensionCapability> headerExtensionsToOffer;
```
Since [array freezing](https://webidl.spec.whatwg.org/#es-frozen-array) is shallow (uses _"[SetIntegrityLevel](https://tc39.es/ecma262/#sec-setintegritylevel)(array, "frozen")"_, same as [Object.freeze](https://tc39.es/ecma262/#sec-object.freeze) —see [fiddle](https://jsfiddle.net/jib1/5dt36qao/)) it only ensures entries cannot be swapped out, but the entries themselves (in this case pojo caches of the dictionary) remain mutable. These mutations are not observable to the user agent, but _are_ observable to JS, which seems confusing. https://github.com/whatwg/webidl/issues/348

henbos:
Completed as part of https://github.com/w3c/webrtc-extensions/pull/142

Issue #132, opened 2022-12-03
Suggestion: rename [[HeaderExtensionsToOffer]] to [[HeaderExtensionsSupported]]


docfaraday:
[[HeaderExtensionsToOffer]] also influences creation of answers, and also cannot force an extension to be put in a reoffer if the previous answer did not have that extension, so having "ToOffer" in its name is a bit of a misnomer.

Also, it probably makes sense to rename the other stuff (eg; setOfferedRtpHeaderExtensions to setSupportedHeaderExtensions), since that is in effect what JS is trying to do.

alvestrand:
Agree that [[HeaderExtensionsToOffer]] needs to be [[HeaderExtensionsSupported]], since it includes the currently "stopped" extensions.
We could define the "header extensions to offer" as the set of header extensions in [[HeaderExtensionsSupported]] that are not currently "stopped" for better language flow. However, "setOfferedRtpHeaderExtensions" does, to my mind, do exactly what it says on the label.

docfaraday:
setOfferedRtpHeaderExtensions strongly influences what we put in both offers and answers, but does not outright dictate what we use. I would argue that it does not do what it says on the tin, with that name. What it does is set the supported RTP extensions. From there, JSEP says we offer the supported RTP extensions minus anything that was not present in the previous answer, and JSEP says we answer with the intersection of the supported RTP extensions and what was present in the offer.

alvestrand:
Hm. I think we're using "supported" in two different fashions here. I was thinking "supported" as "supported in code" - that is, as a synonym for "implemented". I think your use of "supported" is more like "negotiated for use on this transport".

(It's a side effect of the Unified Plan decision that headers are negotiated per media session, but the result has to be consistent across all media sections bundled to a single transport.)

alvestrand:
Perhaps the internal slot is [[HeaderExtensionTargetState]]?
And we could say that setRemoteDescription() flipped the direction of every member that's not present in the remote description to "stopped"?

docfaraday:
I am basing my definition of "supported" on the fact that we will need to define the phrase "supported RTP header extension" in JSEP in terms of this API, if we want to avoid modifying JSEP itself. Generally, over in IETF land, "supported" means "ready and _willing_", so this isn't an unexpected use of terminology.

alvestrand:
[...]

PR #142, opened 2023-02-22
Update header extensions API to get-modify-set.

### LINKED ISSUE #141: Header extension: direction woes and renames
The direction could refer to any of the following:
- What we want to negotiate.
- What we have already negotiated.
- What the user agent uses by default.
- What the user agent is capable of doing.

For example, a header extension might be "stopped" by default, we want to negotiate it as "sendonly" but we are capable of negotiating it as "sendrecv" and at the moment nothing is negotiated, so "stopped".

What's even more confusion about direction is that currently, header extension capabilities are spread out across both `RTCRtpSender.getCapabilities(kind).headerExtensions` and `RTCRtpReciever.getCapabilities(kind).headerExtensions`.

Here's my proposal to reduce confusion and get a more unified API shape, plus some renames while we're at it.
This potentially fixes #130, #132, #137 and #136.

```
partial interface RTCRtpTransciever {
  // Replaces RTCRtpSender/Receiver.getCapabilities(kind).headerExtensions.
  // exts[i].direction refers to what is default, e.g. this could be "stopped" even
  // though it is possible to negotiate it.
  static sequence<RTCRtpHeaderExtensionCapability> getDefaultHeaderExtensions(DOMString kind);

  // Rename of headerExtensionsToOffer, fixes #132.
  // Not using FrozenArray, fixes #137 and #136.
  sequence<RTCRtpHeaderExtensionCapability> getHeaderExtensionsToNegotiate();

  // This gives us the get-modify-set pattern, fixing #130.
  void setHeaderExtensionsToNegotiate(sequence<RTCRtpHeaderExtensionCapability> exts);

  // What has been negotiated already.
  sequence<RTCRtpHeaderExtensionCapability> getNegotiatedHeaderExtensions();
};
```

Here's an example usage:
```
const extensions = transciever.getHeaderExtensionsToNegotiate();
const desiredExtension = extensions.find(ext => ext.uri == 'foo');
if (desiredExtension) {
  // The app should know what directions are applicable to a particular header extension,
  // but if 'sendrecv' isn't applicable, the UA will downgrade this to 'sendonly' or
  // 'recvonly' as necessary. getHeaderExtensionsToNegotiate() will let you know.
  desiredExtension.direction = 'sendrecv';
}
transceiver.setHeaderExtensionsToNegotiate(extensions);

// After an O/A, transceiver.getNegotiatedHeaderExtensions() will match `extensions`
// if the other endpoints accepted everything, otherwise directions may be less.
```

henbos:
Fixes #141, #130, #132, #137, #136.

As per [February Virtual Interim slides](https://docs.google.com/presentation/d/1t9Ijbfjt8c1QAC7G5MrTGjwLzo92VS00ZlwBROAcufc/edit#slide=id.g20f65cf48b6_0_38).

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-extensions/pull/142.html" title="Last updated on Mar 2, 2023, 4:12 PM UTC (9cdffb7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-extensions/142/fd7694d...henbos:9cdffb7.html" title="Last updated on Mar 2, 2023, 4:12 PM UTC (9cdffb7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -127,61 +127,150 @@ partial dictionary RTCRtpHeaderExtensionCapability {
 };
 
 partial interface RTCRtpTransceiver {
-  undefined setOfferedRtpHeaderExtensions(
-sequence&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer);
-  readonly attribute FrozenArray&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsToOffer;
-  readonly attribute FrozenArray&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensionsNegotiated;
+  sequence&lt;RTCRtpHeaderExtensionCapability&gt; getHeaderExtensionsToNegotiate();
+  undefined setHeaderExtensionsToNegotiate(
+      sequence&lt;RTCRtpHeaderExtensionCapability&gt; extensions);
+  sequence&lt;RTCRtpHeaderExtensionCapability&gt; getNegotiatedHeaderExtensions();
 };
     </pre>
     <p>
-      Let <dfn data-dfn-for=RTCRtpTransceiver>[[\HeaderExtensionsToOffer]]</dfn> be an internal slot of the {{RTCRtpTransceiver}},
-      initialized to the platform-specific list of implemented RTP header extensions.
-      The direction
-      attribute for all extensions that are mandatory to use MUST be initialized to
-      an appropriate value other than "stopped". The {{RTCRtpHeaderExtensionCapability/direction}} attribute for extensions
-      that will not be offered by default in an initial offer MUST be initialized to
-      {{RTCRtpTransceiverDirection/"stopped"}}.
+      Let
+      <dfn data-dfn-for=RTCRtpTransceiver>{{RTCRtpTransceiver/[[HeaderExtensionsToNegotiate]]}}</dfn>
+      and
+      <dfn data-dfn-for=RTCRtpTransceiver>{{RTCRtpTransceiver/[[NegotiatedHeaderExtensions]]}}</dfn>
+      be internal slots of the {{RTCRtpTransceiver}}, initialized as follows:
     </p>
-    <p class="note">The list of header extensions that MUST/SHOULD be supported is listed in
-      [[RTCWEB-RTP]], section 5.2. The "mid" extension is mandatory to use when
-      BUNDLE is in use, per [[BUNDLE]] section 9.1.
-    </p>
-    <p>
-      Let <dfn data-dfn-for=RTCRtpTransceiver>{{RTCRtpTransceiver/[[HeaderExtensionsNegotiated]]}}</dfn> be an internal slot of the {{RTCRtpTransceiver}},
-      initialized to an empty list.
-    </p>
+    <ol>
+      <li>
+        <p>Set {{RTCRtpTransceiver/[[HeaderExtensionsToNegotiate]]}} to the
+        platform-specific list of implemented RTP header extensions. The
+        {{RTCRtpHeaderExtensionCapability/direction}} attribute for all
+        extensions that are mandatory to use MUST be initialized to an
+        appropriate value other than {{RTCRtpTransceiverDirection/"stopped"}}.
+        The {{RTCRtpHeaderExtensionCapability/direction}} attribute for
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -191,48 +280,116 @@ <h2>Modifications to existing procedures</h2>
...
             <li>
-              For each extension <var>ext</var> in <var>extlist</var>, do the following:
+              <p>Let <var>transceiver</var> be the {{RTCRtpTransceiver}} that
+              this method was invoked on.</p>
+            </li>
+            <li>
+              <p>Let <var>extensions</var> be the first argument of this
+              method.</p>
+            </li>
+            <li>
```
fippo, line 311: I don't agree with the algorithm here. The following one works nicely and leaves a bit more freedom to developers:
1/ for each extension in HeaderExtensionsToNegotiate set direction to stopped
2/ for each extension in extension find matching extension in HeaderExtensionsToNegotiate and set its direction to extension direction

index.html
```diff
@@ -191,48 +280,116 @@ <h2>Modifications to existing procedures</h2>
...
+                  is {{RTCRtpTransceiverDirection/"stopped"}} and
+                  {{RTCRtpHeaderExtensionCapability/uri}} indicates a
+                  mandatory-to-implement extension, [=exception/throw=] an
+                  {{InvalidModificationError}} and return
+                  <code>undefined</code>.</p>
                 </li>
                 <li>
-                  Set the {{RTCRtpHeaderExtensionCapability/direction}} field of <var>entry</var> to <var>ext</var>'s
-                  {{RTCRtpHeaderExtensionCapability/direction}} field.
+                  <p>If necessary, downgrade
```
fippo, line 363: can you find a better term than "downgrade"?

index.html
henbos: What part of the algorithm do you not agree with? Not being able to re-order or remove things? Why would you ever do that in a "get-modify-set" pattern? What's the benefit of deleting an element over setting the elements direction to stopped?

henbos: I don't want apps to have hardcoded list of header extensions set because it is not very web compat friendly and it ignores browser defaults and implicitly turns off extensions that might be added by default in the future

index.html
fippo:
```suggestion
                  <p>If necessary, restrict
```

henbos: Done

index.html
```diff
@@ -127,61 +127,151 @@ <h3>
...
+        <li>
+          <p>In the steps for processing each <var>media description</var> of a
+          remote <var>description</var> (<var>remote</var> is
+          <code>true</code>), after <var>transceiver</var> has been found or
+          created, insert the following steps:</p>
+          <ol>
+            <li>
+              <p>If <var>description</var> is of type {{RTCSdpType/"offer"}},
+              then for each <var>extension</var> in
+              <var>transciever.</var>{{RTCRtpTransceiver/[[HeaderExtensionsToNegotiate]]}},
```
youennf, line 180:
```suggestion
              <var>transceiver.</var>{{RTCRtpTransceiver/[[HeaderExtensionsToNegotiate]]}},
```

index.html
```diff
@@ -191,48 +281,116 @@ <h2>Modifications to existing procedures</h2>
...
                 <li>
-                  Let <var>entry</var> be the entry with the same {{RTCRtpHeaderExtensionCapability/uri}} value in
-                  {{RTCRtpTransceiver/[[HeaderExtensionsToOffer]]}}.
+                  <p>If
+                  <var>extension</var>.{{RTCRtpHeaderExtensionCapability/uri}}
+                  is not equal to the {{RTCRtpHeaderExtensionCapability/uri}} of
+                  the <var>i</var>-th element of
+                  <var>transceiver</var>.{{RTCRtpTransceiver/[[HeaderExtensionsToNegotiate]]}},
+                  [=exception/throw=] an {{InvalidModificationError}} and return
+                  <code>undefined</code>.</p>
```
youennf, line 342: Why returning undefined after throwing.
We should just abort these steps.

henbos: I'll update to throw

jan-ivar (approved):

henbos:
The checks are complaining about unrelated references to credentials so I'll go ahead and merge

Issue #126, opened 2022-11-30
Negotiationless codec selection and per-encoding layer codec configuration


henbos:
## Problem
With the introduction of [scalabilityMode](https://w3c.github.io/webrtc-svc/#dom-rtcrtpencodingparameters-scalabilitymode) (webrtc-svc spec), we now have two competing APIs for configuring the encoder: setCodecPreferences (or SDP munging) and setParameters() with scalabilityMode. Issues:

1. Different codecs have different scalability modes and toggling "simulcast VP8 without SVC" and "non-simulcast VP9 using SVC" not only changes scalability, it also changes the number of active layers. Not being able to set all settings (=codec, scalability mode, active) in a single API call inevitably leads to temporarily sending the wrong thing. Its racy, causes additional key frames and not very ergonomic.

Other issues not specific to SVC:

2. Some apps don't want to re-negotiate (e.g. fixed number of transceiver setups), besides, why do a round-trip to the server if you already know the server supports the desired codec?

3. An app may want to use different codec configurations for different layers e.g. performance vs bitrate tradeoffs.

## Proposal
Configure send codecs with setParameters(). Add "codec" to [RTCRtpEncodingParameters](https://w3c.github.io/webrtc-pc/#rtcrtpencodingparameters) of type [RTCRtpCodecCapability](https://w3c.github.io/webrtc-pc/#dom-rtcrtpcodeccapability) (same type that is used by getCapabilities and setCodecPreferences).

This means that the codec list order in the SDP will not dictate what codec we use for sending. What about backwards-compatibility? Don't worry, we still default to the first codec in the list, but if the codec value is ever changed, that's what we use going forward.

## Pseudo-PR

Update RTCRtpSender initialization steps:
1. Let sender have a [[PreferSelectedCodecs]] internal slot, initialized to false.
2. Let each encoding in sender's [[SendEncodings]] have a [[SelectedCodec]] internal slot representing the currently selected codec for this encoding, initialized to null.

Update the RTCRtpSender.setParameters() steps:
1. For each encoding in the argument to this function, let codec be the value of encodings[i].codec:
1.1. If codec is not one of the values in RTCRtpSender.getCapabilities(kind).codecs, throw an exception and abort these steps.
1.2. If transceiver.mid is not null and codec is not one of the values in [[SendCodecs]], throw an exception and abort these steps. _(If transceiver.mid is null we've not negotiated yet and allow optimistically selecting one.)_
[...]

PR #147, opened 2023-03-16
Add RTCRtpEncodingParameters.codec to change the active codec

### LINKED ISSUE #43: Mixed-codec simulcast proposal
It would be interesting to give users more control over which codec is used for each simulcast layer.
We used to have `codecPayloadType` that took at payload type (codec id negotiated after O/A) but it was removed as it wasn't well designed (PT not known before O/A) and no implementations supported it.

We have a few use cases in mind where mixed codecs could be used, for example using a lower layer with a higher complexity codec (for example software AV1) to get more efficiency and a higher level with another relatively simpler codec (VP8/9).

I imagine that the each simulcast layer could have an array of codecs, listed in the same way as with `setCodecPreferences()`.

Goals are: 

- Filter out unwanted codecs for specific layers
- Override the priority of the codecs negotiated during the O/A.
- Per-layer control of features like FEC or RTX.
- **CAN'T** add new codecs with this API, we are limited to what is in the O/A.

Let me know if you want a more detailed API proposal with IDL.

Orphis:
This new encoding parameter allows changing the active codec on any simulcast layer as long as it is supported or negotiated.

It allows user agent that don't support the feature partially or fully to reject the configuration change.

Fixes #43 , #126

***
<a href="https://pr-preview.s3.amazonaws.com/Orphis/webrtc-extensions/pull/147.html" title="Last updated on Apr 20, 2023, 2:29 PM UTC (a088a4f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-extensions/147/7f7b5f9...Orphis:a088a4f.html" title="Last updated on Apr 20, 2023, 2:29 PM UTC (a088a4f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -569,7 +569,7 @@ partial interface RTCRtpTransceiver {
         <dd>
           <p>Optional value selecting which codec is used for this encoding's
             RTP stream. The {{RTCRtpCodec}} dictionary is defined in [[WEBRTC]].
-            If absent, the user agent can chose to use any negotiated codec.</p>
+            If [=map/exists|absent=], the user agent can chose to use any negotiated codec.</p>
         </dd>
       </dl>
     </section>
--- a/index.html
+++ b/index.html
@@ -523,6 +523,7 @@ partial interface RTCRtpTransceiver {
     <pre class="idl">partial dictionary RTCRtpEncodingParameters {
   unsigned long ptime;
   boolean adaptivePtime = false;
+  RTCRtpCodec codec;
 };</pre>
     <section id="rtcrtpencodingparameters-attributes">
       <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
@@ -562,7 +563,79 @@ partial interface RTCRtpTransceiver {
           allows the <a>user agent</a> to adapt its bandwidth allocation strategy
           based on the current network conditions.</p>
         </dd>
+        <dt>
+          <dfn data-idl>codec</dfn> of type <span class="idlMemberType">RTCRtpCodec</span>
+        </dt>
+        <dd>
+          <p>Optional value selecting which codec is used for this encoding's
+            RTP stream. The {{RTCRtpCodec}} dictionary is defined in [[WEBRTC]].
+            If absent, the user agent can chose to use any negotiated codec.</p>
+        </dd>
       </dl>
+    </section>
+    <section id="rtcrtpencodingparameters-modifications">
+      <h2>Modifications to existing procedures</h2>
+      <h3>setParameters()</h3>
+      <p>Add the following steps to the [=RTCRtpSender/setParameters validation steps=]:</p>
+      <ol>
+        <li>
+          <p>Let <var>choosableCodecs</var> be parameters.{{RTCRtpParameters/codecs}}.</p>
+        </li>
+        <li>
+          <p>If <var>choosableCodecs</var> is an empty list, set <var>choosableCodecs</var>
+            to transceiver.{{RTCRtpTransceiver/[[PreferredCodecs]]}}.</p>
+        </li>
+        <li>
+            <p>If <var>choosableCodecs</var> is still an empty list, set <var>choosableCodecs</var>
+              to the [=RTCRtpSender/list of implemented send codecs=] for transceiver's kind.</p>
+        </li>
+        <li>
+          <p>If any <var>encoding</var> in encodings [=map/exists|contains=] a codec
+            [= codec match | not found =] in <var>choosableCodecs</var>, return a promise
+            [= rejected =] with a newly [= exception/created =] {{InvalidModificationError}}.</p>
+        </li>
+        <li>
+          <p>If the user agent does not support setting the codec for any encoding or mixing
+            different codec values on the different encodings, return a promise [= rejected =]
+            with a newly [= exception/created =] {{OperationError}}.
+          </p>
+        </li>
+      </ol>
+      <h3>addTransceiver()</h3>
+      <p>Add the following steps to the [=RTCPeerConnection/addTransceiver sendEncodings validation steps=]:</p>
+      <ol>
+        <li>
+          <p>If any <var>codec</var> parameter in <var>sendEncodings</var> does [= codec match | not match =] any codec in
+          {{RTCRtpSender.getCapabilities(kind)}}.<code>codecs</code>,
+          [= exception/throw =] an {{OperationError}}.</p>
+        </li>
+        <li>
+          <p>If the user agent does not support changing codecs without negotiation or
+            does not support setting codecs for individual encodings, return a promise
+            [= rejected =] with a newly [= exception/created =] {{OperationError}}.</p>
+        </li>
+      </ol>
+      <h3>Set the session description</h3>
+      <p>Append the following steps to the <a data-cite="WEBRTC#set-description">set the session description</a> algorithm:</p>
+      <ol>
+        <li>
+          <p>For each <var>transceiver</var> in <var>connection</var>'s [= set of transceivers =]:</p>
+          <ol>
+            <li>
+              <p>Let <var>codecs</var> be <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendCodecs]]}}.</p>
+            </li>
+            <li>
+              <p>If <var>codecs</var> is not an empty list:</p>
+              <ol>
+                <li>
+                  <p>Remove any <var>codec</var> value in <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}
+                    that does not [= codec match | match =] any entry in <var>codecs</var>.</p>
+                </li>
+              </ol>
+            </li>
+          </ol>
+        </li>
+      </ol>
     </section>
   </section>
   <section id="rtcrtpcontributingsource-extensions">
--- a/webrtc-extensions.js
+++ b/webrtc-extensions.js
@@ -5,7 +5,7 @@ var respecConfig = {
       repoURL: "https://github.com/w3c/webrtc-extensions/",
       branch: "main"
     },
-    "xref": ["html", "webidl", "webrtc", "hr-time", "mediacapture-streams", "webrtc-stats"],
+    "xref": ["html", "webidl", "webrtc", "hr-time", "mediacapture-streams", "webrtc-stats", "infra"],
     "shortName": "webrtc-extensions",
     "specStatus": "ED",
     "subjectPrefix": "[webrtc-extensions]",


index.html
```diff
@@ -562,7 +563,75 @@ <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
         allows the <a>user agent</a> to adapt its bandwidth allocation strategy
         based on the current network conditions.</p>
       </dd>
+      <dt><dfn data-idl>codec</dfn> of type <span class="idlMemberType">RTCRtpCodec</span></dt>
+      <dd>
+        <p>Indicates the codec for this encoding. The {{RTCRtpCodec}} dictionary is
```
henbos, line 568: We need to clarify when to use or not use this codec.

For backwards-compat, someone using get/setParameters() should just get "whatever was negotiated", i.e. the order in the SDP.

But if someone says "I want to control codec using this API", I think we need to say to ignore the SDP order, otherwise you would reset the codec in the next SDP negotiation for unrelated reasons.

I suggest the opt-in to control the codec with this API instead of setCodecPreferences() should be "if codec was specified in addTransceiver()". Boolean internal slot for this?

index.html
```diff
@@ -562,7 +563,75 @@ <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
...
+              return a promise [= rejected =] with a newly [= exception/created =]
+              {{InvalidModificationError}}.</p></li>
+        </ol>
+      </li>
+      <li>
+        <p>If the user agent does not support setting the codec for any encoding or mixing
+          different codec values on the different encodings, return a promise [= rejected =]
+          with a newly [= exception/created =] {{OperationError}}.
+        </p>
+      </li>
```
henbos, line 618: In parallel, reconfigure the encoder?

index.html
henbos: If codec is a preference, undefined by default, that solves it too

henbos: How about codecPreferences: undefined by default, once set it overrides the preference order - let's be clear about that

henbos:
Looks great, it just needs some clarifications

index.html
henbos: Luckily addTransceiver() will not acquire any resources, because it hasn't been negotiated yet. The spec currently says to acquire during negotiation, but as pointed out by Bernard in the last Virtual interim meeting, in practise we don't acquire encoder resources until when we actually use them, so any failure would have to be surfaced then - this is an existing problem, and perhaps we'll want to event handlers for this in the future, but that's a separate issue

index.html
```diff
@@ -555,7 +556,102 @@ <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
...
+        <ol>
+          <li>
+            <p>Let <var>codecs</var> be <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendCodecs]]}}.</p>
+          </li>
+          <li>
+            <p>If <var>codecs</var> is not an empty list:</p>
+            <ol>
+              <li>
+                <p>Remove any <var>codec</var> value in <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}
+                  that does not match any entry in <var>codecs</var>.</p>
```
henbos, line 648: Edge case that needs clarification: If a codec that we have previously preferred (encodings[0].codec = "VP8" for example) is removed in response to negotiation, will encodings[0].codec become missing (no preference) or will we continue to prefer VP8 just in case it is negotiated in the future?

index.html
```diff
@@ -555,7 +556,102 @@ <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
...
         based on the current network conditions.</p>
       </dd>
+      <dt><dfn data-idl>codec</dfn> of type <span class="idlMemberType">RTCRtpCodec</span></dt>
+      <dd>
+        <p>Optional value overriding the codec used for this encoding. The {{RTCRtpCodec}} dictionary is
+          defined in [[WEBRTC]]. If absent, the default codec selection algorithm selects the first
+          negotiated codec.</p>
+          <div class="issue atrisk">
+            <p>
+              {{RTCRtpEncodingParameters/codec}} is marked as a feature at risk
```
youennf, line 566: I would not mark it as a feature at risk, maybe mention that it is a new feature asking fro implementers feedback

index.html
Orphis: The goal is to remove the `codec` value override in each entry in `[[SendEncodings]]` that became invalid after negotiation. Do you have a better wording to suggest?

henbos: Nevermind, I was reading this wrong. After re-reading, I think this is clear as-is. LGTM without any changes.

index.html
```diff
@@ -555,7 +556,102 @@ <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
...
+        <ol>
+          <li>
+            <p>If any <var>codec</var> parameter in <var>encodings</var> does not match
+              any entry in <var>codecs</var>, return a promise [= rejected =] with a newly
+              [= exception/created =] {{InvalidModificationError}}.</p>
+          </li>
+        </ol>
+      </li>
+      <li>
+        <p>Else (<var>codecs</var> is an empty list), if <var>transceiver</var>.
```
jan-ivar, line 590: Prose-wise we use "Otherwise," to be less codey — But I find the repeated "Else" pattern here confusing, (even more so with the exists and non-empty). Can we avoid it?

index.html
```diff
@@ -555,7 +556,102 @@ <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
...
+          {{RTCRtpTransceiver/[[PreferredCodecs]]}} is not an empty list:</p>
+        <ol>
+          <li>
+            <p>If any <var>codec</var> parameter in <var>encodings</var> does not match
```
jan-ivar, line 594: Reminder to link to match algorithm if we go with that. And below.

index.html
jan-ivar: > I think the UA should be able to determine ahead of time if it is able to apply this type of configuration, so it should be able to reject it early, without a parallel step.

Depends on your definition of "apply". UAs may be able to reject some combinations early, but it's not clear to me that success is guaranteed any more strongly for customized codecs than it is for default ones.

In any case, while rejecting a promise early has some benefit while stepping through in a JS debugger, there's not much else to gain from that particular timing.

index.html
```diff
@@ -555,7 +556,102 @@ <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
...
+        </p>
+      </li>
+    </ol>
+    <h3>addTransceiver()</h3>
+    <p>Add the following steps in the <var>sendEncodings</var> validation for
+      {{RTCPeerConnection/addTransceiver()}}, before any validation of
+      <var>scalabilityMode</var> from [[WEBRTC-SVC]]:</p>
+    <ol>
+      <li>
+        <p>If any <var>codec</var> parameter in <var>sendEncodings</var> does not match any codec in
```
jan-ivar, line 624: I think we need a check for [map/exists] and a for loop here to spell this out.

index.html
```diff
@@ -555,7 +556,102 @@ <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
...
+        {{RTCRtpSender.getCapabilities(kind)}}.<code>codecs</code>,
+        [= exception/throw =] an {{OperationError}}.</p>
+      </li>
+      <li>
+        <p>If the user agent does not support changing the codec for any encoding or mixing
+          different codec values on the different encodings,
```
jan-ivar, line 630: "mixing different codec values" seems a bit vague. What exactly are the requirements for not throwing OperationError?

index.html
dontcallmedom: marking a feature "at risk" only makes sense in the context of a Candidate Recommendation; since that document is not intended to be published on the Rec track, but instead to feed into the WebRTC Rec (AFAIK), then I don't think marking anything at risk serves much purpose indeed.

index.html
```diff
@@ -555,7 +556,102 @@ <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
         allows the <a>user agent</a> to adapt its bandwidth allocation strategy
         based on the current network conditions.</p>
       </dd>
+      <dt><dfn data-idl>codec</dfn> of type <span class="idlMemberType">RTCRtpCodec</span></dt>
+      <dd>
+        <p>Optional value overriding the codec used for this encoding. The {{RTCRtpCodec}} dictionary is
```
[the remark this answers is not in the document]
henbos: The "negotiation once, never re-negotiate" use case is becoming more popular, where one uses CSRCs to identify which participant is which, but always having a fixed number of transceivers.

When you can do negotiationless codec switching for video, I think it is only a matter of time before there will be an ask for audio as well.

That said the audio use case seems a lot more "at risk" than the video use case. For now I don't mind limiting scope to video-only, but I don't think we should rule out the possibility long term.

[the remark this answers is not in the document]
henbos: One of the bullet points on [this slide](https://docs.google.com/presentation/d/1OfAN3sfVXhvOBCqgLMegrNcHz7-V4qTnuHjBP3gLdgg/edit#slide=id.g216ea3acd8c_1_3) is "make it possible to change codec without re-negotiating". But this API is much more important for video than for audio due to the other two bullet points which are both video-only uses:
- Allow different codecs on different encodings.
- Allow specifying both codec and scalabilityMode with a single API call.

aboba (commented):
The checks for `setParameters()` seem to allow for setting of combinations of `codec` and `scalabilityMode` that are illegal (e.g. setting VP8 and L3T3).  To fix this, I think we need to add a check in WebRTC-SVC.

index.html
aboba: The "feature at risk" designation is there to keep track of what has been implemented (and is therefore eligible to be added to WebRTC-PC).

index.html
```diff
@@ -555,7 +556,102 @@ <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
...
     </dl>
+    <h2>Modifications to existing procedures</h2>
+    <h3>setParameters()</h3>
+    <p>Add the following steps in the parameter validation for {{RTCRtpSender/setParameters()}},
+      before any validation of <var>scalabilityMode</var> from [[WEBRTC-SVC]]:</p>
+    <ol>
+      <li><p>Let <var>codecs</var> be <var> parameters</var>.{{RTCRtpParameters/codecs}}.</p>
+      </li>
+      <li>
+        <p>If <var>codecs</var> is not an empty list:</p>
```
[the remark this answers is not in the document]
Orphis: `codecs` is using `RTCRtpParameters.codecs` which is set to `[[SendCodecs]]`, which is initialized to an empty list. So it will never be missing from the input (other wise, it's a modification of a read only parameter and should be rejected).

index.html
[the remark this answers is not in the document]
Orphis: I don't think we should restrict this API to video only. Codec is a general concept, why should it only apply for video only?
While the slides were presented with examples centered around video, with the emergence of new audio codecs, it seems even more important to keep the symmetry in the APIs.

index.html
Orphis: `OperationError` is already part of the specification when `setParameters()` cannot allocate resources. Chrome is able to throw those when unknown errors occur deep in the pipeline, which we associate with a resource allocation issue for now.
This describes a new behaviour though and I believe returning an `OperationError` in case of a complex unsupported input might be relevant.

henbos (approved):

index.html
```diff
@@ -562,8 +563,120 @@ <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
...
+        </dd>
       </dl>
     </section>
+    <section id="rtcrtpencodingparameters-modifications">
+      <h2>Modifications to existing procedures</h2>
+      <h3>setParameters()</h3>
+      <p>Add the following steps in the parameter validation for {{RTCRtpSender/setParameters()}},
+        before any validation of <var>scalabilityMode</var> from [[WEBRTC-SVC]]:</p>
+      <ol>
+        <li><p>Let <var>codecs</var> be <var> parameters</var>.{{RTCRtpParameters/codecs}}.</p>
```
jan-ivar, line 582: Hmm, webrtc-pc already contains this (unused!) line, so assuming https://github.com/w3c/webrtc-pc/pull/2854 gets merged, how about something like: _"Under the [=setParameters validation steps=]"_, replace this line:

 1. Let codecs be parameters.[codecs](https://w3c.github.io/webrtc-pc/#dom-rtcrtpparameters-codecs).

with:

 1. Let _choosableCodecs_ be parameters.[codecs](https://w3c.github.io/webrtc-pc/#dom-rtcrtpparameters-codecs).
 2. If _choosableCodecs_ is an empty list, set _choosableCodecs_ to <var>transceiver</var>.{{RTCRtpTransceiver/[[PreferredCodecs]]}}
 3. If _choosableCodecs_ is still an empty list, set _choosableCodecs_ to the [=[list of implemented send codecs](https://w3c.github.io/webrtc-pc/#dfn-list-of-implemented-send-codecs)=] for [=[transceiver kind](https://w3c.github.io/webrtc-pc/#dfn-transceiver-kind)=]

And then follow SVC's lead above and say: _"Add the following to the conditions under which the operation causes a promise rejected with an [InvalidModificationError](https://webidl.spec.whatwg.org/#invalidmodificationerror) (step 4 within step 6):"_
- Any _encoding_ in encodings [=map/exists|contains=] a codec [= codec match | not found =] in choosableCodecs

This should avoid the if/else/else and hopefully be readable. WDYT?

index.html
```diff
@@ -562,8 +563,120 @@ <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
...
+            defined in [[WEBRTC]]. If absent, the default codec selection algorithm selects the first
+            negotiated codec.</p>
+        </dd>
       </dl>
     </section>
+    <section id="rtcrtpencodingparameters-modifications">
+      <h2>Modifications to existing procedures</h2>
+      <h3>setParameters()</h3>
+      <p>Add the following steps in the parameter validation for {{RTCRtpSender/setParameters()}},
+        before any validation of <var>scalabilityMode</var> from [[WEBRTC-SVC]]:</p>
```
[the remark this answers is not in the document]
aboba: The problem is that not that the normative reference is incorrect - the problem is that the combined procedure doesn't produce the results you'd expect.

index.html
aboba: To precent attempts to set illegal combinations of codecs and `scalabilityMode` values, I think you need to also set the "most preferred codec" slot and refer to this slot in WebRTC-SVC Section 4.2.2.

aboba (changes requested):
I don't think that the proposed checks work together with the WebRTC-SVC checks to produce the expected results.  See other comments.

index.html
Orphis: Feature at risk has been removed. We can add it later if implementations are lacking.

jan-ivar (commented):
whoops, had this pending comment from before that was not submitted

PR #160, opened 2023-04-19
Change playoutDelay to jitterBufferTarget.

### LINKED ISSUE #12: Testability of playoutDelayHint

### LINKED ISSUE #46: How does a developer decide on a value for `playoutDelay` ?
In general, APIs that take a numerical value for a tradeoff are bad, because it's hard to determine the threshold values between the various use-cases. How does a developer find what the best value is for a particular use-case?

If the best value for a use-case is fixed (which seems to be the case, looking at https://github.com/w3c/webrtc-extensions/issues/8), then an enum is better. If not, then other APIs that allow determining this value for a particular environment must be available.

### LINKED ISSUE #123: RTCRtpReceiver.playoutDelay...Hint or no Hint?
According WPT test status, all browsers have implemented playoutDelayHint:
https://wpt.fyi/results/webrtc-extensions/RTCRtpReceiver-playoutDelayHint.html?label=experimental&label=master&aligned

But the spec says [playoutDelay](https://w3c.github.io/webrtc-extensions/#dom-rtcrtpreceiver-playoutdelay) ever since [this PR](https://github.com/w3c/webrtc-extensions/pull/22/files)... ironically the referenced issue talks about not wanting to go back and forth on the name, but here we are.

Seems like we should update the spec to call it what the implementations do?

### LINKED ISSUE #155: playoutDelay should use milliseconds.
W3C design principle [§ 8.3. Use milliseconds for time measurement](https://w3ctag.github.io/design-principles/#milliseconds) says... well, exactly that.

Let's not spread the mistake from stats of using seconds (which I feel personally responsible for) to more specs!

Since no browser implements this feature correctly yet — fortunately, Chrome only implements a cousin named `playoutDelayHint` that fails to throw on input validation — I still see an opportunity to correct this.

### LINKED ISSUE #156: Should we clarify playoutDelay value is jitter buffer depth?
https://github.com/w3c/webrtc-extensions/issues/46#issuecomment-661779681 and https://github.com/w3c/webrtc-extensions/issues/12 suggest a shared understanding that `playoutDelay` controls the user agent's jitter buffer depth, and little else. They stress the importance of JS being able to read the UA's current value, both for tests and correct API use.

This depth is JS-observable through `receiver.getStats()`, as experiments in https://github.com/w3c/webrtc-extensions/issues/12#issuecomment-1497681908 show.

But [current spec language](https://w3c.github.io/webrtc-extensions/#dom-rtcrtpreceiver-playoutdelay) instead ties it to _"time between network reception of media and playout."_ i.e. target jitter buffer depth + implementation-defined playout path delay.

Ironically, I think this stems from early concerns over web compatibility (including mine): Naively it seemed better to me if the same value in two browsers produced the same net total playout delay (the "result"). But hindsight, that "result" isn't just total playout delay, but more or less jitter!

I think applications using this API ultimately wish to set the jitter buffer depth directly. 

While the spec already alludes to this in notes, "[notes in this specification are non-normative](https://w3c.github.io/webrtc-extensions/#conformance)":

image

I think we should update the normative language to clarify that the value is the target jitter buffer depth and nothing else.

### LINKED ISSUE #157: Does playoutDelay of video affect jitter buffer of synced audio and vice versa?
The spec has [this non-normative note](https://w3c.github.io/webrtc-extensions/#dfn-maximum-allowed-delay):
 
image

What does _"the largest of those hints will take precedence in synchronization mechanism"_ mean? Does it mean:
1. tracks get synced at playout without the benefit of deeper jitter buffers?
2. tracks get the same (larger) jitter buffer depth?

As mentioned in #156, jitter buffer depth is JS-observable through receiver.getStats(), so we should be clear.

Notes are [non-normative](https://w3c.github.io/webrtc-extensions/#conformance) in this spec, and I find no normative steps to support 2, which suggests 1.

Was 1 the intent? Is 2 useful? If so, do we wish to specify it? Any downsides to doing so? Either way we should clarify.

jan-ivar:
Fixes #156.
Fixes #157.
Fixes #155.
Fixes #123.
Fixes #46.
Fixes #12.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-extensions/pull/160.html" title="Last updated on Apr 20, 2023, 8:05 PM UTC (bc36d73)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-extensions/160/fa46b8c...jan-ivar:bc36d73.html" title="Last updated on Apr 20, 2023, 8:05 PM UTC (bc36d73)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -359,7 +359,7 @@ partial interface RTCRtpTransceiver {
     <section id="rtcrtpreceiver-jitterbuffertarget-rtcrtpreceiver-interface">
       <h2>{{RTCRtpReceiver}} interface extensions</h2>
       <pre class="idl">partial interface RTCRtpReceiver {
-  attribute DOMHighResTimestamp? jitterBufferTarget;
+  attribute DOMHighResTimeStamp? jitterBufferTarget;
 };</pre>
     </section>
     <section id="rtcrtpreceiver-jitterbuffertarget-attributes">
@@ -367,7 +367,7 @@ partial interface RTCRtpTransceiver {
       <dl data-link-for="RTCRtpReceiver" data-dfn-for="RTCRtpReceiver"
       class="attributes">
         <dt><dfn>jitterBufferTarget</dfn> of type <span
-          class="idlAttrType">DOMHighResTimestamp</span>, nullable</dt>
+          class="idlAttrType">DOMHighResTimeStamp</span>, nullable</dt>
         <dd>
           <p>This attribute allows the application to specify a target duration
           of time in milliseconds of media for the {{RTCRtpReceiver}}'s jitter
--- a/index.html
+++ b/index.html
@@ -359,7 +359,7 @@ partial interface RTCRtpTransceiver {
     <section id="rtcrtpreceiver-jitterbufferdepth-rtcrtpreceiver-interface">
       <h2>{{RTCRtpReceiver}} interface extensions</h2>
       <pre class="idl">partial interface RTCRtpReceiver {
-  attribute double? jitterBufferDepth;
+  attribute DOMHighResTimestamp? jitterBufferDepth;
 };</pre>
     </section>
     <section id="rtcrtpreceiver-jitterbufferdepth-attributes">
--- a/index.html
+++ b/index.html
@@ -13,6 +13,7 @@
       This document defines a set of ECMAScript APIs in WebIDL to extend the WebRTC 1.0 API.
     </p>
   </section>
+  </section>
   <section id="sotd">
     <p>The API is based on preliminary work done in the W3C WEBRTC Working Group.</p>
   </section>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -339,73 +340,62 @@ <h3>Methods</h3>
...
+          of time in milliseconds of media for the {{RTCRtpReceiver}}'s jitter
+          buffer to hold. This influences the amount of buffering done by the
+          <a>user agent</a>, and allows applications to control the tradeoff between
+          playout delay and the risk of running out of audio or video frames due to
+          network jitter.
+
+          <p>The <a>user agent</a> MUST have a <dfn>minimum allowed depth</dfn> and a
+          <dfn>maximum allowed depth</dfn> reflecting what the <a>user agent</a> is
           able or willing to provide based on network conditions and memory
           constraints.</p>
```
henbos, line 382:
```suggestion
          constraints, and may change at any time.</p>
```

henbos: Just to clarify that these are not constant min and max values, if that wasn't already obvious

index.html
```diff
@@ -339,73 +340,62 @@ <h3>Methods</h3>
...
+          <div class="note">
+            <p>This is a target value. The resulting change in delay can be gradually
+            observed over time. The receiver's average jitter buffer delay can be
+            measured as the delta
             {{RTCInboundRtpStreamStats/jitterBufferDelay}} divided by the delta
             {{RTCInboundRtpStreamStats/jitterBufferEmittedCount}}.
-          </p>
+            </p>
+            <p>
+            An average delay is expected even if DTX is used. For example, if
```
henbos, line 391: What's DTX?

index.html
```diff
@@ -414,37 +404,50 @@ <h2>Attributes</h2>
...
+              to <var>depth</var>.</p>
             </li>
             <li>
-              <p>If <var>delay</var> is negative or larger than 4 seconds then,
-              [=exception/throw=] a {{RangeError}} and abort these steps.</p>
+              <p>If <var>depth</var> is below <a>minimum allowed
+              depth</a> or above <a>maximum allowed depth</a> then clamp
+              <var>depth</var> to <a>minimum allowed
+              depth</a> or <a>maximum allowed depth</a> to be as close as
+              possible to the original value.</p>
```
henbos, line 422: Since min and max is something that changes continuously, it would be wrong to apply the clamping as a one-time event. If min is 1 ms most of the time, but then there is a sudden glitch and min is momentarely set to 1000 ms, but the user sets jitterBufferDepth to 50 ms, I would expect the jitter buffer target to be as close to 50 ms as possible as soon as "min" has recovered, and not forever be stuck at 1000 ms as this is not what the app asked for.

Also doing this in the JS thread doesn't make sense since the jitter buffer min and max values would adjusted by the user agent in parallel.

I have a suggested edit below to replace this paragraph.

henbos, line 422:
```suggestion
```

index.html
```diff
@@ -414,37 +404,50 @@ <h2>Attributes</h2>
...
-                  value used MUST be clamped to <a>minimum allowed
-                  delay</a> or <a>maximum allowed delay</a> to be as close as
-                  possible to the requested one.</p>
+                </li>
+                <p>
+                  If <var>track</var> is synchronized with another
+                  {{RTCRtpReceiver}}'s track for
+                  <a data-cite="RFC5888#section-7">audio/video synchronization</a>,
+                  then the <a>user agent</a> SHOULD use the larger of the two receivers'
+                  {{RTCRtpReceiver/[[JitterBufferDepth]]}} for both receivers.
```
henbos, line 441:
```suggestion
                  {{RTCRtpReceiver/[[JitterBufferDepth]]}} for both receivers.
                </p>
                <p>
                  When the underlying system is applying a jitter buffer depth, it will
                  continuously make sure that the actual jitter buffer depth is clamped
                  within the <a>minimum allowed depth</a> and <a>maximum allowed
                  depth</a>.
```

henbos (changes requested):

murillo128:
question: is the jitterBufferDepth also meant to increase the RTX retrasmission timers (i.e. the amount of time the "jitter buffer" will be sendind nacks for a missing packet?)

henbos:
I don't have insight into how the RTX timers work, but my naive assumption is that WebRTC should do sensible things related to what the size of the jitter buffer is, so why should it matter _why_ the jitter buffer increased. The API is just one thing that affects jitter buffer, but there are other things too. Nothing special about this API in that sense

index.html
```diff
@@ -339,73 +340,62 @@ <h3>Methods</h3>
...
+        Let {{RTCRtpReceiver}} objects have a <dfn data-dfn-for=RTCRtpReceiver>[[\JitterBufferDepth]]</dfn> internal slot initially
         initialized to <code>null</code>.
       </p>
     </section>
-    <section id="rtcrtpreceiver-playoutdelay-rtcrtpreceiver-interface">
+    <section id="rtcrtpreceiver-jitterbufferdepth-rtcrtpreceiver-interface">
       <h2>{{RTCRtpReceiver}} interface extensions</h2>
       <pre class="idl">partial interface RTCRtpReceiver {
-  attribute double? playoutDelay;
+  attribute double? jitterBufferDepth;
```
henbos, line 362: Should we use DOMHighResTimestamp instead? It's also a double but used for milliseconds and this is ms now.

> The DOMHighResTimeStamp type is a double and is used to store a time value in milliseconds

index.html
jan-ivar: Ah, that was not clear before, so thanks for clarifying! We should avoid MAY here then probably 
```suggestion
          constraints, which can change at any time.</p>
```

henbos: sgtm

index.html
jan-ivar: It's [existing text](https://w3c.github.io/webrtc-extensions/#dom-rtcrtpreceiver-playoutdelay). I believe it's https://www.rfc-editor.org/rfc/rfc4788#section-1.3

index.html
jan-ivar: Yes, me moving these out was indeed based on my assumption they were constant. Note that even [in parallel](https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel) steps _"are to be run, one after another"_ in sequence, so this wasn't clear before. I know we sometimes use a shorthand to describe future reactions inside steps, when _"When X happens, do this"_ seems too much.

index.html
henbos: Ack

index.html
jan-ivar: We could, but it's not a timestamp, so that seems confusing. It's only mentioned under [§ 8.4. Use the appropriate type to represent times and dates](https://w3ctag.github.io/design-principles/#times-and-dates).

index.html
fippo: https://www.rfc-editor.org/rfc/rfc7587#section-3.1.3 is a better reference.

index.html
henbos, line 362: jitterBufferTarget?

henbos (approved):
Approve % "DOMHighResTimeStamp?" and name jitterBufferTarget

aboba:
@jan-ivar  Can you fix the "[[ReceiverTrack]]" issue?

bdrtc:
> > I don't have insight into how the RTX timers work, but my naive assumption is that WebRTC should do sensible things related to what the size of the jitter buffer is, so why should it matter why the jitter buffer increased. The API is just one thing that affects jitter buffer, but there are other things too. Nothing special about this API in that sense
> 
> Given that we have changed the `playoutDelay` to the `jitterbufferDepth` I think it would be good to reference that the increase of the buffer should also affect to the retransmissions and packet loss recovery mechanism in place, if not increasing the buffer depth has only effects in controlling the playout delay and I would prefer to stick to the old name.


rtx-time was added in webrtc 111 release, when setting jitterbufferTarget, user may change rtx-time together via sdp.
this is [ rtx-time issue](https://bugs.chromium.org/p/webrtc/issues/detail?id=12420)

Issue #134, opened 2022-12-04
Should setOfferedHeaderExtensions be allowed to restart stopped extensions?


alvestrand:
Spinoff from #132 since we should decide what makes sense.

JSEP section 5.2.2 says that subsequent offer/answer can only use the set of header extensions from the initial offer/answer - no adding header extensions later.

We modify (in section 6.1) the procedures for subsequent offers indicating that it is OK to turn on header extensions later.
If this is needed in a realistic scenario, we should support it; if it's not needed, there's no real reason for changing JSEP on this point.

Possible approaches are:

1. Allow turning on header extensions later (overriding current JSEP text)
2. Let SetOfferedHeaderExtensions throw an InvalidStateError if the negotiation state is not "new" and the call attempts to change an extensions that is not in [[HeaderExtensionsNegotiated]] from "stopped" to something else
3. Let the algorithm in section 6.1 iterate over [[HeaderExtensionsNegotiated]] rather than over [[HeaderExtensionsToOffer]], effectively following the current JSEP rule, but providing no feedback to the user that he attempted something impossible

My current thinking is that I think favor approach \#2, possibly also doing \#3. I'm sure there are other possible approaches.

fippo:
How does this work currently for setCodecPreferences?
From JSEP wrt media formats:
> and MUST include all currently available formats
but indeed for header extensions:
> The RTP header extensions MUST only include those that are present in the most recent answer.

One of the arguments I made for not needing support for `enabled` was that you can just renegotiate any header extensions. As far as I can see,
https://jsfiddle.net/fippo/6waoyLqs/1/
reoffers ssrc-audio-level in Chrome and Firefox.

fippo:
IMO we should specify what browsers currently support, try to negotiate header extensions again and let O/A figure it out

fippo:
Note that we already do modify JSEP [here](https://w3c.github.io/webrtc-extensions/#modifications-to-existing-procedures-0)

henbos:
The spec already says...

[...]

PR #164, opened 2023-04-26
Remove JSEP Modifications

aboba:
Fixes https://github.com/w3c/webrtc-extensions/pull/134

There is an existing process for submitting RFC 8829 errata, as well as for updating RFC 8829bis (now in the RFC Editor queue).  Let's see if we can address these JSEP issues some other way.  Justin Uberti will join us at the May 4 WEBRTC WG Editor's meeting to discuss alternatives.

***
<a href="https://pr-preview.s3.amazonaws.com/aboba/webrtc-extensions/pull/164.html" title="Last updated on May 16, 2023, 9:45 AM UTC (d2869bd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-extensions/164/51024d3...aboba:d2869bd.html" title="Last updated on May 16, 2023, 9:45 AM UTC (d2869bd)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -207,9 +207,9 @@ partial interface RTCRtpTransceiver {
         </li>
       </ul>
     <p>
-      In [[JSEPbis]] Sections 5.2.1 and 5.2.2 an RTP header extension "e" is considered "enabled" if it is
-      listed in {{RTCRtpTransceiver/[[HeaderExtensionsToNegotiate]]}} where {{RTCRtpHeaderExtensionCapability/direction}} is not 
-      {{RTCRtpTransceiverDirection/"stopped"}}.
+      In [[JSEPbis]] Section 5.2.1 (Initial Offers) and 5.2.2 (Subsequent Answers) an RTP header extension "e" is
+      considered "enabled" if it is listed in {{RTCRtpTransceiver/[[HeaderExtensionsToNegotiate]]}} where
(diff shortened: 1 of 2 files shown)


henbos:
JSEP talks about "For each supported RTP header extension", so one might argue that a JSEP-compatible version of this PR that doesn't turn the API into a NO-OP might be to replaces the JSEP modifications with one new step at the end of `setHeaderExtensionsToNegotiate`: "Modify the internal set of supported RTP header extensions to match [[HeaderExtensionsToNegotiate]]."

You might want to lawyer whether that is or is not OK but it seems better than a NO-OP until we have something in JSEP to do what we want.

aboba:
@henbos RFC 8829bis assumes that "supported" RTP header extensions are always offered, and if offered, are answered. But with the RTP header extension API only the "mandatory to use" RTP header extensions are automatically offered/answered;  a "supported" RTP header extension also needs to be in the [[HeaderExtensionsToNegotiate]].  So if we can still change RFC 8829bis, perhaps we can swap out the word "supported" for something better without having to refer to internal slots. IF we can't modify RFC 8829bis, your suggestion might work.

henbos:
https://github.com/rtcweb-wg/jsep/issues/1032

alvestrand:
In order to not derail our work that depends on the JSEP changes, I suggest that we update the JSEP reference to point to the -bis draft as part of this PR.
If we can't say "JSEP needs to change" in our document, we need to be able to point to the WIP document that makes the changes.

PR #168, opened 2023-05-25
Prevent candidate pair removal

### LINKED ISSUE #166: ICE improvements: prevent candidate pair removal
**Background**: An incremental approach for improving ICE control capabilities was discussed at the WebRTC WG [April 2023 interim meeting](https://www.w3.org/2023/04/18-webrtc-minutes.html#t08). This issue relates to the first in a series of proposed improvements.

**Problem**: Applications should be able to maintain multiple candidate pairs by preventing the removal of candidate pairs that are not in active use for transport. This is necessary for applications that want to maintain redundant candidate pairs for connection resiliency, or move traffic to an alternate route based on the network cost or other factors.

**Proposal**: This could be achieved a few different ways.

1.  Cancelable event for candidate pair removal

    When the ICE agent has decided to prune a candidate pair (eg. due to inactivity), the user agent fires a [`cancelable`](https://dom.spec.whatwg.org/#ref-for-dom-event-cancelable%E2%91%A1) event to let the application know that a candidate pair is about to be removed. The application can prevent the removal from happening by calling [`preventDefault()`](https://dom.spec.whatwg.org/#ref-for-dom-event-preventdefault%E2%91%A2) on the event. If not [prevented](https://dom.spec.whatwg.org/#ref-for-dom-event-defaultprevented%E2%91%A0), the ICE agent is free to prune the candidate pair.
  
    The API looks like
  
    ```webidl
    partial interface RTCIceTransport  {
      attribute EventHandler /* RTCIceCandidatePairEvent */ oncandidatepairadded;      // not strictly necessary for this to work
      attribute EventHandler /* RTCIceCandidatePairEvent */ oncandidatepairremoval;    // before removal
    };
    
    interface RTCIceCandidatePairEvent : Event {    // cancelable
      readonly attribute RTCIceCandidatePair candidatePair;
    };
    ```

    This approach to preventing the user agent from taking default action has a precedent in [touch event](https://w3c.github.io/touch-events/#cancelability) and form [submit event](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#submitevent).

    This event conveys to the application exactly when the removal is to occur, and allows the decision to permit or prevent removal to be deferred until it actually needs to be made. It sidesteps any race conditions that may have to be averted with the other options. 

2.  Change automatic behaviour - `removable` attribute

    The application can change the ICE agent's automatic behaviour by setting an attribute (or an internal slot, through a method) to prevent a candidate pair from being automatically pruned by the ICE agent.
  
    ```webidl
    partial interface RTCIceTransport  {
      attribute EventHandler /* RTCIceCandidatePairEvent */ oncandidatepairadded;
      attribute EventHandler /* RTCIceCandidatePairEvent */ oncandidatepairremoved;    // after removal
    };
    
    partial interface RTCIceCandidatePair {
      attribute boolean removable;
    };
    ```

3. Change automatic behaviour - `idleTimeout` attribute

    This attribute leaves it up to the ICE agent to manage the candidate pair lifecycle, but with input from the application. The ICE agent may prune the candidate pair if no data or ICE check activity has occurred on the candidate pair within the timeout duration.

    ```webidl
    partial interface RTCIceTransport  {
      attribute EventHandler /* RTCIceCandidatePairEvent */ oncandidatepairadded;
      attribute EventHandler /* RTCIceCandidatePairEvent */ oncandidatepairremoved;    // after removal
    };
    
    partial interface RTCIceCandidatePair {
      attribute unsigned long idleTimeout;    // initialized to the default timeout
    };
    ```

    The `idleTimeout` can achieve the same effect as `removable`, with max value meaning never remove, and 0 meaning the ICE agent can prune freely.

    This option gives more control than a `removable` attribute - an application can choose to simply extend the duration for which an inactive candidate pair is retained.

    Another benefit is that there is no need for a new API to let the application prune candidate pairs. The ICE agent will eventually clean up inactive candidate pairs after the timeout. The user agent could also enfore policies on how long of a timeout the application is allowed to set.

sam-vi:
Fixes #166 

This is a proposal for a cancelable event approach to preventing ICE candidate pair pruning.

An event is also proposed at the formation of candidate pairs, which is not strictly necessary for preventing candidate pair pruning, but makes sense to go together for applications to have a better view of the candidate pair lifecycle.

***
<a href="https://pr-preview.s3.amazonaws.com/sam-vi/webrtc-extensions/pull/168.html" title="Last updated on Sep 4, 2023, 2:22 PM UTC (9c3d3f0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-extensions/168/9601ac8...sam-vi:9c3d3f0.html" title="Last updated on Sep 4, 2023, 2:22 PM UTC (9c3d3f0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -750,22 +750,22 @@ dictionary RTCEncodingOptions {
     </ol>
     <pre class="idl">
       partial interface RTCIceTransport {
-        attribute EventHandler oncandidatepairadded;
+        attribute EventHandler onicecandidatepair;
         attribute EventHandler onicecandidatepairprune;
       };</pre>
     <section>
       <h2>Attributes</h2>
       <dl data-link-for="RTCIceTransport" data-dfn-for="RTCIceTransport" class="attributes">
         <dt>
-          <dfn>oncandidatepairadded</dfn> of type <span class="idlAttrType">{{EventHandler}}</span>
+          <dfn>onicecandidatepair</dfn> of type <span class="idlAttrType">{{EventHandler}}</span>
         </dt>
         <dd>
           <p>
-            The event type of this event handler is {{candidatepairadded}}.
+            The event type of this event handler is {{icecandidatepair}}.
           </p>
           <p>
             When the [= ICE agent =] has formed a candidate pair, the [= user agent =] MUST queue a task to [= fire an
-            event =] named {{candidatepairadded}} using the {{RTCIceCandidatePairEvent}} interface with the
+            event =] named {{icecandidatepair}} using the {{RTCIceCandidatePairEvent}} interface with the
             {{RTCIceCandidatePairEvent/candidatePair}} attribute set to the formed candidate pair.
           </p>
         </dd>
@@ -788,7 +788,7 @@ dictionary RTCEncodingOptions {
         <dfn>RTCIceCandidatePairEvent</dfn>
       </h2>
       <p>
-        The {{RTCIceTransport/candidatepairadded}} and {{RTCIceTransport/icecandidatepairprune}} events use the {{RTCIceCandidatePairEvent}} interface.
+        The {{RTCIceTransport/icecandidatepair}} and {{RTCIceTransport/icecandidatepairprune}} events use the {{RTCIceCandidatePairEvent}} interface.
       </p>
       <div>
         <pre class="idl">[Exposed=Window]
@@ -1265,7 +1265,7 @@ partial interface RTCRtpTransceiver {
       </thead>
       <tbody>
         <tr>
-          <th scope=row><dfn data-dfn-for="RTCIceTransport" data-dfn-type=event>candidatepairadded</dfn></th>
+          <th scope=row><dfn data-dfn-for="RTCIceTransport" data-dfn-type=event>icecandidatepair</dfn></th>
           <td>{{RTCIceCandidatePairEvent}}</td>
           <td>
             The [= ICE agent =] has formed a candidate pair and is making it available to the script.
--- a/index.html
+++ b/index.html
@@ -57,9 +57,9 @@
       <a href="https://webrtc.org/experiments/rtp-hdrext/abs-capture-time"><dfn>RTP Header Extension for Absolute Capture Time</dfn></a>:
     </p>
     <ul>
-      <li><a href="https://webrtc.org/experiments/rtp-hdrext/abs-capture-time#absolute-capture-timestamp"><dfn>absolute capture timestamp</dfn></a>
-      <li><a href="https://webrtc.org/experiments/rtp-hdrext/abs-capture-time#timestamp-interpolation"><dfn>timestamp interpolation</dfn></a>
-      <li><a href="https://webrtc.org/experiments/rtp-hdrext/abs-capture-time#estimated-capture-clock-offset"><dfn>estimated capture clock offset</dfn></a>
+      <li><a href="https://webrtc.org/experiments/rtp-hdrext/abs-capture-time#absolute-capture-timestamp"><dfn>absolute capture timestamp</dfn></a></li>
+      <li><a href="https://webrtc.org/experiments/rtp-hdrext/abs-capture-time#timestamp-interpolation"><dfn>timestamp interpolation</dfn></a></li>
+      <li><a href="https://webrtc.org/experiments/rtp-hdrext/abs-capture-time#estimated-capture-clock-offset"><dfn>estimated capture clock offset</dfn></a></li>
     </ul>
     <p>The process of <dfn>chaining</dfn> an operation to an <dfn>operations chain</dfn> is defined in [[WEBRTC]] Section 4.4.1.2.</p>
     <p>
--- a/index.html
+++ b/index.html
@@ -733,8 +733,7 @@ dictionary RTCEncodingOptions {
         <p>
           Let |accepted:boolean| be the result of [= fire an event | firing an event =] named
           {{RTCIceTransport/icecandidatepairprune}} at |transport|, using {{RTCIceCandidatePairEvent}}, with the
-          {{Event/cancelable}} attribute initialized to <code>true</code>, and the {{RTCIceCandidatePairEvent/candidatePair}}
-          attribute initialized to |candidatePair|.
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -604,6 +616,149 @@ <h3>Set the session description</h3>
...
+        <dfn>RTCIceCandidatePairEvent</dfn>
+      </h2>
+      <p>
+        The {{RTCIceTransport/candidatepairadded}} and {{RTCIceTransport/candidatepairpruneproposed}} events use the {{RTCIceCandidatePairEvent}} interface.
+      </p>
+      <div>
+        <pre class="idl">[Exposed=Window]
+interface RTCIceCandidatePairEvent : Event {
+  constructor(DOMString type, RTCIceCandidatePairEventInit eventInitDict);
+  readonly attribute RTCIceCandidatePair candidatePair;
```
youennf, line 717: We usually do not introduce dictionary attribute.
We could expose local and remote directly.

youennf: Another approach discussed during the call: add a RTCIceCandidatePair interface

henbos: The issue is that an attribute cannot be a dictionary, so it needs to return an interface-version of the candidate pair or contain references to the local and remote directly

index.html
```diff
@@ -604,6 +616,149 @@ <h3>Set the session description</h3>
...
+      <li>
+        <p>
+          Otherwise, instruct the [= ICE agent =] to prune the candidate pair indicated by |candidatePair|.
+        </p>
+      </li>
+    </ol>
+    <pre class="idl">
+      partial interface RTCIceTransport {
+        attribute EventHandler oncandidatepairadded;
+        attribute EventHandler oncandidatepairpruneproposed;
```
jan-ivar, line 674: I think we should simplify and call it `oncandidatepairprune`. E.g. we don't say `onclickproposed`.

index.html
sam-vi: I agree with making RTCIceCandidate and RTCIceCandidatePair consistent by adding an interface for the latter.

The current dictionary could be changed to an interface like such:
```webidl
interface RTCIceCandidatePair {
    readonly attribute RTCIceCandidate? local;
    readonly attribute RTCIceCandidate? remote;
};
```

Potentially breaking changes due to switching RTCIceCandidatePair from a dictionary to an interface would be
- an ordered map like `{ local: <candidate>, remote: <candidate> }` could no longer be treated implicitly as an RTCIceCandidatePair
- setting `local` or `remote` on an RTCIceCandidatePair instance will fail on the interface

Is there a precedent for a change like this? Or is the suggestion to add a completely new interface by a different identifier?

index.html
```diff
@@ -604,6 +616,149 @@ <h3>Set the session description</h3>
...
+                The {{candidatePair}} attribute represents the {{RTCIceCandidatePair}} object associated with the event.
+              </p>
+            </dd>
+          </dl>
+        </section>
+      </div>
+      <div>
+        <pre class="idl">
+dictionary RTCIceCandidatePairEventInit : EventInit {
+  required RTCIceCandidatePair candidatePair;
```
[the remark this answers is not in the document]
sam-vi: Done, unless we do add an interface for `RTCIceCandidatePair`.

I've kept `RTCIceCandidate` in the Init rather than `RTCIceCandidateInit`, consistent with [`RTCPeerConnectionIceEventInit`](https://www.w3.org/TR/webrtc/#dom-rtcpeerconnectioniceeventinit).

index.html
sam-vi: I've switched to local+remote `RTCIceCandidate` for now. We can continue the discussion in the June interim or here.

sam-vi: Another option from https://www.w3.org/TR/api-design/#attribute-constant-or-exception-fields - use type `any` for `candidatePair` attribute, and state in the description that `candidatePair` is of type `RTCIceCandidatePair`.

index.html
```diff
@@ -604,6 +616,166 @@ <h3>Set the session description</h3>
...
+      </li>
+      <li>
+        <p>
+          Otherwise, instruct the [= ICE agent =] to prune the candidate pair indicated by |candidatePair|.
+        </p>
+      </li>
+    </ol>
+    <pre class="idl">
+      partial interface RTCIceTransport {
+        attribute EventHandler onicecandidatepair;
```
pthatcher, line 672: It seems like these should be symmetric, like onicecandidateadd/onicecanddiateremove.  

I also like "remove" more than "prune".

pthatcher (approved):
Other than the name change I suggested, I like this PR.

index.html
sam-vi: `onicecandidate` was what I was trying to imitate. But after some thought I'm coming around to `onicecandidatepairadd`/`onicecandidatepairremove`.

"Prune" has a specific, and rather different, meaning in the ICE spec (https://www.rfc-editor.org/rfc/rfc8445#section-6.1.2.4). To avoid any confusion, it makes sense to use the more neutral term "remove". In which case I prefer the symmetry of add/remove despite the departure from `onicecandidate`.

Now to make the add/remove events actually symmetric, there should be an `onicecandidatepairremove` event for every discarded candidate pair which was previously announced with `onicecandidatepairadd`, whether the discard was a result of failed connectivity checks or nomination of a different (higher-priority) candidate pair.

I previously considered separate events, i.e. `onicecandidatepair[delete|remove]` and `onicecandidatepairprune`, for these two scenarios, but it might make sense to bundle those into this single event. The event is cancelable if the application has a choice over the removal, otherwise it is not cancelable and calling `preventDefault()` has no effect.

The only scenario when a **_cancelable_** `onicecandidatepairremove` is fired is:
- another candidate pair is nominated for this data stream, and this candidate pair is not currently used for transport
  - calling `preventDefault()` retains the candidate pair for future selection, and connectivity checks continue

This is the also the only scenario addressed by the PR in its current form.

In all other scenarios, a **_non-cancelable_** `onicecandidatepairremove` is fired, eg.:
- connectivity checks fail for this candidate pair
- the network interface associated with the active candidate pair goes away
- this candidate pair was retained previously by calling `preventDefault()` on the event, is not in active use for transport, and the associated network interface goes away
- this candidate pair is not in active use for transport, was retained by the local peer, but not retained by the remote peer, the associated remote candidate is freed, and connectivity check subsequently fails
- removeCandidatePairs(...) is called for this candidate pair.

alvestrand:
Is this one ready to merge?

henbos (approved):

jan-ivar (approved):

Issue #170, opened 2023-06-12
ICE improvements: remove candidate pairs


sam-vi:
**Background**: Issue #166 proposes a mechanism with which applications can prevent candidate pairs from being pruned by the ICE agent. This allows applications to maintain multiple candidate pairs for connection redundancy and quick switchover.

**Problem**: The ICE agent continues to send connectivity checks on unpruned candidate pairs. This consumes resources on the peers, eg. open ports, power consumption on network interfaces that could've been in sleep, and network capacity for the checks themselves. Resources are also needed on intermediate network hops to maintain unpruned pairs, eg. open NAT bindings and TURN allocations.

Applications should be able to trigger the pruning of candidate pairs when no longer necessary.

**Proposal**: Introduce a new API to allow applications to instruct the ICE agent to prune candidate pairs

```webidl
partial interface RTCIceTransport {
  undefined pruneCandidatePairs(sequence<RTCIceCandidatePair> pairs);
}
```

Calling the method causes the ICE agent to immediately prune all candidate pairs in the method's argument without generating a prune proposal. No further connectivity checks are sent or answered on the pruned pairs.

The method fails with an error if called in an invalid context (eg. from an event handler for prune proposal) or with an invalid argument (deleted or non-existent candidate pairs).

If the currently selected candidate pair for transport is pruned by the application in this way, this is treated in the same way as if the selected pair was destroyed, triggering the selection of a different candidate pair, or an ICE failure if no other viable pair can be found.

PR #175, opened 2023-09-05
Add RTCIceTransport method to remove candidate pairs

sam-vi:
Fixes w3c/webrtc-extensions#170.

This PR extends the `RTCIceTransport` interface by adding a new method to allow an application to remove candidate pairs that are no longer needed for the peer connection. The new method `removeCandidatePairs` immediately removes the supplied candidate pairs, which are then sent in non-cancelable `icecandidatepairremove` events.

***
<a href="https://pr-preview.s3.amazonaws.com/sam-vi/webrtc-extensions/pull/175.html" title="Last updated on Jan 12, 2024, 10:42 AM UTC (d0d0ae1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-extensions/175/48e2a4a...sam-vi:d0d0ae1.html" title="Last updated on Jan 12, 2024, 10:42 AM UTC (d0d0ae1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -855,7 +855,7 @@ dictionary RTCEncodingOptions {
           attribute EventHandler onicecandidatepairremove;
           attribute EventHandler onicecandidatepairnominate;
           undefined setSelectedCandidatePair(RTCIceCandidatePair candidatePair);
-          Promise&lt;undefined&gt; removeCandidatePairs(sequence&lt;RTCIceCandidatePair&gt; candidatePairs);
+          Promise&lt;undefined&gt; removeCandidatePair(RTCIceCandidatePair candidatePair);
         };</pre>
     <section id="rtcicetransport-attributes">
       <h2>Attributes</h2>
@@ -969,11 +969,11 @@ dictionary RTCEncodingOptions {
           </p>
         </dd>
         <dt>
-          <dfn>removeCandidatePairs</dfn>
+          <dfn>removeCandidatePair</dfn>
         </dt>
         <dd>
           <p>
-            The {{removeCandidatePairs}} method removes the provided candidate pairs. The [= ICE agent =] will stop sending and responding to ICE connectivity checks on the removed candidate pairs, and they can no longer be used to send data for this transport. This method is meant to be called when the application wants to free resources allocated for candidates that it no longer needs. The application cannot call {{RTCIceTransport/setSelectedCandidatePair}}() with a candidate pair removed with this method.
+            The {{removeCandidatePair}} method removes the provided candidate pair. The [= ICE agent =] will stop sending and responding to ICE connectivity checks on the removed candidate pair, and it can no longer be used to send data for this transport. This method is meant to be called when the application wants to free resources allocated for candidates that it no longer needs. The application cannot call {{RTCIceTransport/setSelectedCandidatePair}}() with a candidate pair removed with this method.
           </p>
           <p>
             When this method is invoked, the [= user agent =] MUST run the following steps:
@@ -1004,12 +1004,12 @@ dictionary RTCEncodingOptions {
             </li>
             <li>
               <p>
-                Let |candidatePairs:sequence&lt;RTCIceCandidatePair&gt;| be the method's first argument.
+                Let |candidatePair:RTCIceCandidatePair| be the method's first argument.
               </p>
             </li>
             <li>
               <p>
-                Let |filteredCandidatePairs:sequence&lt;RTCIceCandidatePair&gt;| be the result of filtering down <var>candidatePairs</var> to just the {{RTCIceCandidatePair}} elements whose  {{RTCIceCandidatePair/local}} and {{RTCIceCandidatePair/remote}} attributes respectively match a pairing of {{RTCIceCandidatePairEvent/local}} and {{RTCIceCandidatePairEvent/remote}} sent in {{RTCIceTransport/onicecandidatepairadd}}.
+                If the {{RTCIceCandidatePair/local}} and {{RTCIceCandidatePair/remote}} attributes of <var>candidatePair</var> do not match a pairing of {{RTCIceCandidatePairEvent/local}} and {{RTCIceCandidatePairEvent/remote}} respectively sent in {{RTCIceTransport/onicecandidatepairadd}}, [= exception/throw =] a {{NotFoundError}}.
               </p>
             </li>
             <li>
@@ -1019,7 +1019,7 @@ dictionary RTCEncodingOptions {
             </li>
             <li>
               <p>
-                In parallel, instruct the [= ICE agent =] to remove the candidate pairs indicated by <var>filteredCandidatePairs</var>.
+                In parallel, instruct the [= ICE agent =] to remove the candidate pair indicated by <var>candidatePair</var>.
               </p>
               <ol>
                 <li>
@@ -1029,7 +1029,7 @@ dictionary RTCEncodingOptions {
                   <ol>
                     <li>
                       <p>
-                        For each |removedPair:RTCIceCandidatePair| in <var>filteredCandidatePairs</var> that was removed by the [= ICE agent =], [= fire an event =] named {{RTCIceTransport/icecandidatepairremove}} at |transport|, using {{RTCIceCandidatePairEvent}}, with the {{Event/cancelable}} attribute initialized to <code>false</code>, and the {{RTCIceCandidatePairEvent/local}} and {{RTCIceCandidatePairEvent/remote}} attributes initialized to the {{RTCIceCandidatePair/local}} and {{RTCIceCandidatePair/remote}} candidates, respectively, of <var>removedPair</var>.
+                        [= Fire an event =] named {{RTCIceTransport/icecandidatepairremove}} at |transport|, using {{RTCIceCandidatePairEvent}}, with the {{Event/cancelable}} attribute initialized to <code>false</code>, and the {{RTCIceCandidatePairEvent/local}} and {{RTCIceCandidatePairEvent/remote}} attributes initialized to the {{RTCIceCandidatePair/local}} and {{RTCIceCandidatePair/remote}} candidates, respectively, of <var>candidatePair</var>.
                       </p>
                     </li>
                     <li>
--- a/index.html
+++ b/index.html
@@ -1001,7 +1001,7 @@ dictionary RTCEncodingOptions {
         </dt>
         <dd>
           <p>
-            The {{removeCandidatePair}} method removes the provided candidate pair. The [= ICE agent =] will stop sending and responding to ICE connectivity checks on the removed candidate pair, and it can no longer be used to send data for this transport. This method is meant to be called when the application wants to free resources allocated for candidates that it no longer needs. The application cannot call {{RTCIceTransport/setSelectedCandidatePair}}() with a candidate pair removed with this method.
+            The {{removeCandidatePair}} method removes the provided candidate pair. The [= ICE agent =] will stop sending and responding to ICE connectivity checks on the removed candidate pair, and it can no longer be used to send data for this transport. This method is meant to be called when the application wants to free resources allocated for candidates that it no longer needs. The application cannot call {{RTCIceTransport/selectCandidatePair}}() with a candidate pair removed with this method.
           </p>
           <p>
             When this method is invoked, the [= user agent =] MUST run the following steps:
--- a/index.html
+++ b/index.html
@@ -985,8 +985,7 @@ dictionary RTCEncodingOptions {
         </dt>
         <dd>
           <p>
-            The {{removeCandidatePairs}} method removes the provided candidate pairs, making them unavailable to use
-            for transport. The [= ICE agent =] may free up candidates once all the associated candidate pairs have been removed.
+            The {{removeCandidatePairs}} method removes the provided candidate pairs. The [= ICE agent =] will stop sending and responding to ICE connectivity checks on the removed candidate pairs, and they can no longer be used to send data for this transport. This method is meant to be called when the application wants to free resources allocated for candidates that it no longer needs. The application cannot call {{RTCIceTransport/setSelectedCandidatePair}}() with a candidate pair removed with this method.
           </p>
           <p>
             When this method is invoked, the [= user agent =] MUST run the following steps:
@@ -994,7 +993,7 @@ dictionary RTCEncodingOptions {
           <ol class="algorithm">
             <li>
               <p>
-                Let |connection:RTCPeerConnection| be the {{RTCPeerConnection}} object associated with this [= ICE agent =].
+                Let |connection:RTCPeerConnection| be the {{RTCPeerConnection}} object associated with [= this =].
               </p>
             </li>
             <li>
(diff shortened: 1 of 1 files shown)


aboba:
@pthatcher PTAL

henbos:
FYI reviewers, the PR is now rebased and much smaller than it looked before

index.html
```diff
@@ -975,6 +976,67 @@ <h2>Methods</h2>
             selected candidate pair =].
           </p>
         </dd>
+        <dt>
+          <dfn>removeCandidatePairs</dfn>
+        </dt>
+        <dd>
+          <p>
+            The {{removeCandidatePairs}} method immediately removes the provided candidate pairs, making them unavailable to use
```
henbos, line 984: I would avoid words like "immediately" since the ICE agent lives on a different thread, unless "instruct the ICE agent" necessarily needs to be implemented synchronously. But my understanding is that "instruct" could be implemented as "post a task to instruct"... but see my other comment, it is basically the same thing

index.html
```diff
@@ -975,6 +976,67 @@ <h2>Methods</h2>
...
+            </li>
+            <li>
+              <p>
+                Instruct the [= ICE agent =] to remove the candidate pairs indicated by |candidatePairs|, ignoring any unknown
+                or invalid candidate pairs.
+              </p>
+            </li>
+            <li>
+              <p>
+                For each valid |removedPair| in |candidatePairs| that was removed by the [= ICE agent =], [= fire an event =] named
```
henbos, line 1030: It sounds like to know which pairs was removed here, we would need to block JS on ICE agent work.

Should this be in a separate task, i.e. "When the ICE agent as removed all candidate pairs, queue a task that runs the following steps: for each |removePair| in |removedCandidatePairs|, fire ..."

henbos (commented):
I overall like this, just have a question about the "immediate-ness" of this operation

index.html
sam-vi: Ack, removed "immediately" to avoid any confusion.

index.html
sam-vi: Right, we shouldn't block JS on the ICE agent's work. Moved the events firing to a separate task.

Also changed removeCandidatePairs to return a promise that resolves after all changes occurring due to the method have taken effect. This follows a similar pattern for batched actions elsewhere, eg. setRemoteDescription, i.e. return a promise, do everything that has an effect, fire all events, finally resolve the promise.

henbos: Excellent!

henbos (approved):
LGTM

index.html
```diff
@@ -975,6 +980,86 @@ <h2>Methods</h2>
             selected candidate pair =].
           </p>
         </dd>
+        <dt>
+          <dfn>removeCandidatePairs</dfn>
+        </dt>
+        <dd>
+          <p>
+            The {{removeCandidatePairs}} method removes the provided candidate pairs, making them unavailable to use
+            for transport. The [= ICE agent =] may free up candidates once all the associated candidate pairs have been removed.
```
pthatcher, line 989: I'm not sure the last statement is always correct.  Won't a web app be able to keep a local candidate alive even if there are no candidate pairs at the moment?

pthatcher:
Seems fine to me except for the sentence I mentioned.

index.html
sam-vi: The current extent of the (merged) API only allows the application to keep candidate pairs alive, not the candidates themselves. We can revisit the statement once the scope expands to candidate manipulation as well.

index.html
```diff
@@ -825,6 +825,11 @@ <h3>
           {{RTCIceTransport/onicecandidatepairadd}}, [= exception/throw =] a {{NotFoundError}}.
         </p>
       </li>
+      <li>
+        <p>
+          If |candidatePair| describes a [= candidate pair removed =] by the [= ICE agent =] or removed by the application by calling {{RTCIceTransport/removeCandidatePairs}}, [= exception/throw =] a {{NotFoundError}}.
```
jan-ivar, line 830: _candidatePair_ is a dictionary, so identity-matching won't work as input validation. Inputs can be composed pairs that never existed before, e.g. `{local, remote}` where `local` and `remote` objects are from different pairs.

Options:
1. Make [RTCIceCandidatePair](https://w3c.github.io/webrtc-pc/#dom-rtcicecandidatepair) an interface (may be reasonable if remixing pairs is undesirable)
2. Define a comparison algorithm for pairs (e.g. `a.local == b.local && a.remote == b.remote`)

There's also a question of whether `a.local == b.local` is an identity comparison or if two different instances of [RTCIceCandidate](https://w3c.github.io/webrtc-pc/#dom-rtcicecandidate) with identical members are equal (I suspect that's what we want, since the interface is JSON stringifiable).

Also, hunting down what happened to previous pairs doesn't seem like a practical implementation strategy. We probably want something like:
- If candidatePair doesn't "match"(-ing algorithm) an entry in [[ExposedCandidatePairs]]... and then maintain that internal slot.
-

index.html
```diff
@@ -975,6 +980,86 @@ <h2>Methods</h2>
...
+              </p>
+              <ol>
+                <li>
+                  <p>
+                    When the [= ICE agent =] has completed the removal, [= queue a task =] to run the following steps:
+                  </p>
+                  <ol>
+                    <li>
+                      <p>
+                        For each valid |removedPair| in |candidatePairs| that was removed by the [= ICE agent =], [= fire an event =] named {{RTCIceTransport/icecandidatepairremove}}
```
jan-ivar, line 1043: "valid" seems insufficiently clear. 1) what does it mean, and 2) what happens to invalid ones? skipped? Should we instead have synchronous input validation steps that reject?

index.html
```diff
@@ -916,6 +920,7 @@ <h3>
           attribute EventHandler onicecandidatepairremove;
           attribute EventHandler onicecandidatepairnominate;
           undefined setSelectedCandidatePair(RTCIceCandidatePair candidatePair);
+          Promise&lt;undefined&gt; removeCandidatePairs(sequence&lt;RTCIceCandidatePair&gt; candidatePairs);
```
jan-ivar, line 923: Why does it take an array?

What's wrong with:
```js
await Promise.all(pairs.map(pair => transceiver.transport.iceTransport.removeCandidatePair(pair)));
```

index.html
```diff
@@ -975,6 +980,86 @@ <h2>Methods</h2>
             selected candidate pair =].
           </p>
         </dd>
+        <dt>
+          <dfn>removeCandidatePairs</dfn>
+        </dt>
+        <dd>
+          <p>
+            The {{removeCandidatePairs}} method removes the provided candidate pairs, making them unavailable to use
```
jan-ivar, line 988: What's the use case for this method? What happens when the selected pair is removed?

index.html
```diff
@@ -975,6 +980,86 @@ <h2>Methods</h2>
...
+              </p>
+            </li>
+            <li>
+              <p>
+                Let |p:Promise| be a new promise.
+              </p>
+            </li>
+            <li>
+              <p>
+                In parallel, instruct the [= ICE agent =] to remove the candidate pairs indicated by |candidatePairs|, ignoring any unknown or invalid candidate pairs.
```
jan-ivar, line 1033: What's the requirement on the ICE agent here? What does "remove" mean in the context of the ICE agent? Is it a ban against adding it again later?

jan-ivar (changes requested):
What's the use case that requires this new method?

index.html
jan-ivar: A "pair" is just a dictionary containing two [RTCIceCandidate](https://w3c.github.io/webrtc-pc/#dom-rtcicecandidate) objects, which in turn are [not backed by any network resources](https://w3c.github.io/webrtc-pc/#dfn-creating-an-rtcicecandidate). They can be created out of thin air, and thus don't hold anything alive.

Also, let's avoid lower-case "may" and observational statements in general in normative prose.

User agents generally need requirements NOT to save resources. Is this trying to say:
```suggestion
            for transport. The [= ICE agent =] MUST NOT free up candidates until all their associated candidate pairs have been removed.
```
?

jan-ivar: It's also unclear what "free up candidates" means. It seems to refer to the underlying concept "candidate" in [RFC8445](https://www.rfc-editor.org/rfc/rfc8445#section-2.1), not the [RTCIceCandidate](https://w3c.github.io/webrtc-pc/#dom-rtcicecandidate) WebIDL API object.

Do we need to say anything at all?

jan-ivar: Ah "free candidates" is https://datatracker.ietf.org/doc/html/rfc8445#section-8.3. We should probably clarify that with a link if we need this at all.

index.html
sam-vi: I've filed #186 and #187 for matching candidates and pairs respectively.

Matching against previously seen pairs here is drafted similar to [getLocalCandidates](https://w3c.github.io/webrtc-pc/#dom-rtcicetransport-getlocalcandidates), which is implicitly tracked.

But given that inputs must be synchronously validated against the previously seen pairs, I've also filed #188 to add an internal slot.

index.html
sam-vi: The array input is meant as an optimization for a few reasons.

Removing candidate pairs can potentially generate a lot of activity:
- a remove event for each removed pair
- state change events if the currently selected candidate pair is removed

With array, there is also the possibility for an application to select a candidate pair and then remove all unused pairs in a single operation. In general, removing several pairs can be expected to be a common operation when many remove events are cancelled and/or after the application selects a pair for good.

index.html
sam-vi: Replaced the observational statement with the effects and the intended use of the method instead.

index.html
sam-vi: Added a filtering step. Matching to be addressed with https://github.com/w3c/webrtc-extensions/issues/186 and https://github.com/w3c/webrtc-extensions/issues/187.

index.html
sam-vi: Clarified the effects and intended usage of the method.

index.html
sam-vi: "Remove" stops connectivity checks and prevents the use of the candidate pair for sending data after removal. This is similar to the ICE agent discarding duplicate and lower priority pairs. The removed pairs cannot be added back again until candidates are gathered again, which is only possible today with an ICE restart.

index.html
jan-ivar: > The array input is meant as an optimization for a few reasons.

What needs to be optimized?

> Removing candidate pairs can potentially generate a lot of activity:
> 
>     * a remove event for each removed pair

We either want the events or not, and if we do, we should fire them individually I think, and not entertain an (array) API that might get us into considering coalescing events on a single task. I think that only complicates matters and makes this harder for apps to reason about.

If we don't want events, maybe we shouldn't fire them for JS-initiated actions? (I believe I've questioned this elsewhere).

>     * state change events if the currently selected candidate pair is removed

There's only one selected pair, so I don't see what an array optimizes. 

> With array, there is also the possibility for an application to select a candidate pair and then remove all unused pairs in a single operation. In general, removing several pairs can be expected to be a common operation when many remove events are cancelled and/or after the application selects a pair for good.

Wht does "for good" mean? Is it
1. let nomination complete, or
2. remove all pairs but the selected pair?

If it's 2 then what happens to connectivity checks? Do they continue, wasting resources?

I'm concerned this API is a workaround for the "never nominate" hack that might leave apps to different ways to conclude ICE, with one draining more resources than the other.

index.html
jan-ivar: Can we link to prose in [RFC8445](https://www.rfc-editor.org/rfc/rfc8445) that allow user agents to modify its operation in this way?

index.html
sam-vi: > If we don't want events, maybe we shouldn't fire them for JS-initiated actions?

We want all events to fire for a consistent API - all candidate pair additions are announced by the `icecandidatepairadd` event and removals announced by the `icecandidatepairremove` event, and always after the ICE agent has updated its internal state. I think selectively not firing a subset of events would make the API clunkier, even if all capability was retained (of which I am not completely certain in any case).

> If it's 2 then what happens to connectivity checks? Do they continue, wasting resources?

Removing candidate pairs will stop connectivity checks and release their resources. That is the explicit goal of this API and I've made edits to clarify that in the latest commit.

> I'm concerned this API is a workaround for the "never nominate" hack that might leave apps to different ways to conclude ICE, with one draining more resources than the other.

Giving apps more control over ICE is the central mission of the new APIs. Yes, that might mean using more resources if, for instance, the app thinks it will improve the reliability of the call. An app could also drain resources by creating tons of other peer connections, so I'm not sure I follow how this API makes it possible to drain more resources than many other options that are already possible today.

Having said that, changing the API to take a single argument instead of an array doesn't make any functional difference to the capability of the API. I can change it to a single argument and we could revisit and add an array overload in the future if it seems useful.

index.html
sam-vi: "Remove" isn't explicitly defined as a concept in RFC8445. It is something the ICE agent just does during the course of another operation:

1. [Section 6.1.2.4](https://www.rfc-editor.org/rfc/rfc8445#section-6.1.2.4) Pruning Pairs: The agent prunes each checklist.  This is done by **removing** a candidate pair if it is redundant with a higher-priority candidate pair in the same checklist.

2. [Section 6.1.2.5](https://www.rfc-editor.org/rfc/rfc8445#section-6.1.2.5) Removing Lower-Priority Pairs: The limit is enforced by, within in each checklist, **discarding** lower-priority candidate pairs...

So how about this as the requirement for the ICE agent:
1. Remove the provided candidate pair from all Checklists in which it is present.
2. [Update the states](https://www.rfc-editor.org/rfc/rfc8445#section-7.2.5.4) of the affected Checklists.
3. If any local candidates are now unused because they are no longer paired with any remote candidates, the ICE agent can [free those candidates](https://www.rfc-editor.org/rfc/rfc8445#section-8.3).

This means removed candidate pairs are not added back (at least not until regathering is possible).

henbos:
Is this ready to merge at the next meeting?

jan-ivar (approved):
Overall I don't mind this API, so I'm approving it for the sake of progress.

However, I worry there are some API races here that we still need to sort out. We appear to have two async methods, select and remove basically, which appear as if they can race. We should clarify what's supposed to happen then.

For instance, what happens if someone calls both methods at once with the same argument?

My recommendation of adding an internal slot in a follow-up might help clarify this.

index.html
```diff
@@ -1022,6 +1027,90 @@ <h2>Methods</h2>
...
+              </p>
+            </li>
+            <li>
+              <p>
+                Let |candidatePair:RTCIceCandidatePair| be the method's first argument.
+              </p>
+            </li>
+            <li>
+              <p>
+                If the {{RTCIceCandidatePair/local}} and {{RTCIceCandidatePair/remote}} attributes of <var>candidatePair</var> do not match a pairing of {{RTCIceCandidatePairEvent/local}} and {{RTCIceCandidatePairEvent/remote}} respectively sent in {{RTCIceTransport/onicecandidatepairadd}}, [= exception/throw =] a {{NotFoundError}}.
```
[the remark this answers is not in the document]
jan-ivar: @sam-vi if you can file a new issue on this when you merge this PR that would be great thanks!

index.html
```diff
@@ -1022,6 +1027,90 @@ <h2>Methods</h2>
...
+        <dt>
+          <dfn>removeCandidatePair</dfn>
+        </dt>
+        <dd>
+          <p>
+            The {{removeCandidatePair}} method removes the provided candidate pair. The [= ICE agent =] will stop sending and
+            responding to ICE connectivity checks on the removed candidate pair, and it can no longer be used to send data for this
+            transport. This method is meant to be called when the application wants to allow the [= ICE agent =] to [= free =]
+            candidates that it no longer needs. The application cannot call {{RTCIceTransport/selectCandidatePair}}() with a
+            candidate pair removed with this method.
```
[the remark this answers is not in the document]
sam-vi: Removed the text.

Issue #188 addresses the requirement on the user agent.

PR #194 fixes it by doing two things:
1. Maintains an internal slot of _valid_ candidate pairs.
2. Validates the input to `selectCandidatePair` against the slot.

I will send #194 for review once the prerequisite candidate pair match algorithm in PR #193 is merged.

index.html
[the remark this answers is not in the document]
sam-vi: Yes, good point. The validation should certainly be specified better.

Issue  #188 covers input validation of `removeCandidatePair` against the internal slot.

I have drafted the validation in my fork in sam-vi/webrtc-extensions#4, awaiting PRs with the prerequisites to be merged.

index.html
```diff
@@ -1022,6 +1027,90 @@ <h2>Methods</h2>
...
+              </p>
+              <ol>
+                <li>
+                  <p>
+                    When the [= ICE agent =] has completed the removal, [= queue a task =] to run the following steps:
+                  </p>
+                  <ol>
+                    <li>
+                      <p>
+                        [= Fire an event =] named {{RTCIceTransport/icecandidatepairremove}} at |transport|, using {{RTCIceCandidatePairEvent}}, with the {{Event/cancelable}} attribute initialized to <code>false</code>, and the {{RTCIceCandidatePairEvent/local}} and {{RTCIceCandidatePairEvent/remote}} attributes initialized to the {{RTCIceCandidatePair/local}} and {{RTCIceCandidatePair/remote}} candidates, respectively, of <var>candidatePair</var>.
```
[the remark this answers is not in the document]
sam-vi: Ack, happy to discuss this separately. Perhaps after the validations that are in-flight are merged, we can see if something needs further tightening.

index.html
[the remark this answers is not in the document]
sam-vi: > Also, is JS allowed to remove the currently selected candidate pair? What happens then?

Yes. This is not prevented by any current or planned validations.

This should trigger the [`disconnected`](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceTransport/state#the_disconnected_state) state, similar to the loss of the active candidate pair:

> Every [user agent](https://developer.mozilla.org/en-US/docs/Glossary/User_agent) and platform may have its own scenarios that can trigger the "disconnected" state. Possible causes include:
>    - The network interface being used by the connection has gone offline.

sam-vi:
> Overall I don't mind this API, so I'm approving it for the sake of progress.
> 
> However, I worry there are some API races here that we still need to sort out. We appear to have two async methods, select and remove basically, which appear as if they can race. We should clarify what's supposed to happen then.
> 
> For instance, what happens if someone calls both methods at once with the same argument?
> 
> My recommendation of adding an internal slot in a follow-up might help clarify this.

Thank you!

I think the internal slot and validations should address a lot of the race conditions. I would certainly want to fix any remaining issues once the proposed changes are submitted. I am happy to continue discussing and iterating.

henbos:
OK let's merge this, @sam-vi vill you care a follow-up issue regarding internal slots and validation?

PR #177, opened 2023-10-27
Don't fire icecandidatepairadd if closed + avoid competing normative prose

### LINKED ISSUE #176: icecandidatepairadd is fired on closed pc
The steps to fire `icecandidatepairadd` are located under [attributes](https://w3c.github.io/webrtc-extensions/#rtcicetransport-attributes), and fail to check [[isClosed]].

There's also inconsistent use of _"When the ICE agent [does something"_ normative statements in this  section.

It might be better to move these statements out and avoid some of the competing normative statements.

jan-ivar:
Fixes #176. cc @sam-vi — these are some items I missed in review.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-extensions/pull/177.html" title="Last updated on Oct 27, 2023, 8:07 PM UTC (44c10dd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-extensions/177/2833e5b...jan-ivar:44c10dd.html" title="Last updated on Oct 27, 2023, 8:07 PM UTC (44c10dd)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -75,6 +75,10 @@
       The terms [= event =], [= event handlers =] and [= event handler event types =] are defined in [[!HTML]].
     </p>
     <p>
+      The process of <dfn data-lt="form|formed|forming">forming</dfn> a candidate pair is defined in
+      [[RFC8445]] Section 6.1.2.2.
+    </p>
+    <p>
       The process of <dfn data-lt="nominate|nominated|nomination">nominating</dfn> a candidate pair is defined in
       [[RFC8445]] Section 8.1.1.
     </p>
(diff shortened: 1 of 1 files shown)


sam-vi:
LGTM, makes sense to guard the event. Thanks!

index.html
```diff
@@ -717,6 +721,41 @@ <h3>
       Changing
       the selected candidate pair after a successful nomination requires an ICE restart.
     </p>
+    <p>
+      When the [= ICE agent =] has [= formed =] a candidate pair, the [= user agent =] MUST [= queue a task =] to <dfn
```
pthatcher, line 725: I think it's a little vague *when* a candidate pair is "formed".  So I think it might be better to phrases this in terms of the ICE agent is going to add a candidate pair, and just avoid the use of the term "form".

sam-vi: I take it you are referring to [Section 6.1.2.2](https://www.rfc-editor.org/rfc/rfc8445#section-6.1.2.2), where candidate pairs are "formed", but some of these can be discarded in sections [6.1.2.3](https://www.rfc-editor.org/rfc/rfc8445#section-6.1.2.3), [6.1.2.4](https://www.rfc-editor.org/rfc/rfc8445#section-6.1.2.4) and [6.1.2.5](https://www.rfc-editor.org/rfc/rfc8445#section-6.1.2.5) before any connectivity checks are ever sent. 

Only at the end of these steps does the ICE agent have a list of "added" candidate pairs on which connectivity checks could be sent, and only these bear surfacing up to the application. RFC 8445 does not have any terminology for candidate pairs that get to this stage.

So perhaps the definition in terminology could be rephrased as:

"The process of `adding` candidate pairs is defined as forming candidate pairs and pruning redundant pairs in [RFC 8445] Section [6.1.2](https://www.rfc-editor.org/rfc/rfc8445#section-6.1.2)."

henbos:
Editors can integrate if @jan-ivar files an issue

index.html
henbos: Merging this one, feel free to file an issue @pthatcher if you think this can be improved

PR #181, opened 2023-10-30
Move algorithm to setSelectedCandidatePair method, and use [=this=].

### LINKED ISSUE #180: What happens if setSelectedCandidatePair() is called early?
In [§ 9. RTCIceTransport extensions](https://w3c.github.io/webrtc-extensions/#rtcicetransport) between two _"When the [ICE agent](https://w3c.github.io/webrtc-extensions/#dfn-ice-agent) [does X]"_ sections, it says:
- _"If the application defers the [nomination](https://w3c.github.io/webrtc-extensions/#dfn-nominate) of a candidate pair by cancelling the [icecandidatepairnominate](https://w3c.github.io/webrtc-extensions/#dfn-icecandidatepairnominate) event, it may select a different candidate pair to send data by calling [setSelectedCandidatePair](https://w3c.github.io/webrtc-extensions/#dom-rtcicetransport-setselectedcandidatepair)."_

We should reserve "may" and "MAY" for implementations and avoid using it to describe requirements on applications.

The intent of this statement is also unclear, since an application can call a method at any time. If we meant for this method to throw or reject if called outside this condition, that needs to go into its algorithm.

Also, the [change the selected candidate pair](https://w3c.github.io/webrtc-extensions/#rtcicetransport-select) algorithm itself has some bugs, like inferring _transport_ from the passed-in candidatePair.

I think it's better to move this algorithm to the method section.

Assuming there is no limit on when this method can be called, I'll do a PR to that effect.

jan-ivar:
Fixes #180. cc @sam-vi

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-extensions/pull/181.html" title="Last updated on Oct 30, 2023, 10:23 PM UTC (21756cf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-extensions/181/2833e5b...jan-ivar:21756cf.html" title="Last updated on Oct 30, 2023, 10:23 PM UTC (21756cf)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -779,66 +779,6 @@ dictionary RTCEncodingOptions {
       </li>
     </ol>
     <p>
-      If the application defers the {{nomination}} of a candidate pair by cancelling the
-      {{RTCIceTransport/icecandidatepairnominate}} event, it may select a different candidate pair to send data by calling
-      {{RTCIceTransport/setSelectedCandidatePair}}. When the method is called, the [= user agent =] MUST run the steps to <dfn
-        id="rtcicetransport-select">change the selected candidate pair</dfn>:
-    </p>
-    <ol class="algorithm">
-      <li>
-        <p>
-          Let |connection:RTCPeerConnection| be the {{RTCPeerConnection}} object associated with this [= ICE agent =].
-        </p>
-      </li>
-      <li>
-        <p>
-          If <var>connection</var>.{{RTCPeerConnection/[[IsClosed]]}} is
-          <code>true</code>, [= exception/throw =] an {{InvalidStateError}}.
-        </p>
-      </li>
-      <li>
-        <p>
-          Let |transport:RTCIceTransport| be the {{RTCIceTransport}} object associated with this candidate pair.
-        </p>
-      </li>
-      <li>
-        <p>
-          If |transport|.{{RTCIceTransport/[[ProposalPending]]}} is <code>true</code>, [= exception/throw =] an {{InvalidStateError}}.
-        </p>
-      </li>
-      <li>
-        <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -970,9 +910,62 @@ <h2>Methods</h2>
...
+            It is meant to be called after the application defers the {{nomination}} of a candidate pair
+            by cancelling the {{RTCIceTransport/icecandidatepairnominate}} event.
+          <p>
+          <p>
+            When this method is invoked, the [= user agent =] MUST run the following steps:
+          </p>
+          <ol class="algorithm">
+            <li>
+              <p>
+                Let |connection:RTCPeerConnection| be the {{RTCPeerConnection}} object associated with [=this=].
```
henbos, line 924: nit: "associated with this" sounds like we're referring to something close to "this" but other than "this", when we are in fact referring to "this". Should this be "Let |connection:RTCPeerConnection| be [=this=] {{RTCPeerConnection}}."?

Or if [=this=] is already a variable you can refer to, we don't need |connection|, we can just s/connection/[=this=]

henbos (approved):
LGTM % nit, unless @sam-vi had a plan to re-use this algorithm in other places too

index.html
henbos: Sorry ignore my comment we're not on the peer connection!

sam-vi:
LGTM, thanks for the improvements!

pthatcher:
LGTM

alvestrand (approved):
I have an outstanding question about why we can't do SelectCandidatePair while ProposalPending is true, but that's not changed in this PR. Approving.

Issue #188, opened 2023-11-06
Maintain an internal slot of candidate pairs that are valid inputs to RTCIceTransport operations


sam-vi:
An internal slot of [[ActiveCandidatePairs]] can track candidate pairs that are
- added when sent in onicecandidatepairadd
- removed when sent in onicecandidatepairremove, which is not canceled
- removed when supplied to removeCandidatePairs

This makes it explicit and possible to validate synchronously whether an input to an RTCIceTransport operation is valid.

jan-ivar:
It might make sense to build this on top of https://github.com/w3c/webrtc-pc/issues/2906.

PR #194, opened 2024-01-10
Add an internal slot for active candidate pairs.

sam-vi:
Fixes: #188.

The [[CandidatePairs]] slot in RTCIceTransport tracks candidate pairs formed by the ICE agent and surfaced to the application, and not removed via icecandidatepairremove or removeCandidatePair().

The slot is used to validate the inputs to selectCandidatePair() and removeCandidatePair().

**EDIT:** To help Sameer make progress while OOO, I've re-uploaded this as #197 with comments addressed.

***
<a href="https://pr-preview.s3.amazonaws.com/sam-vi/webrtc-extensions/pull/194.html" title="Last updated on Jan 25, 2024, 12:54 PM UTC (0c3c86c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-extensions/194/4c29534...sam-vi:0c3c86c.html" title="Last updated on Jan 25, 2024, 12:54 PM UTC (0c3c86c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1090,7 +1090,12 @@ dictionary RTCEncodingOptions {
             </li>
             <li>
               <p>
-                If the {{RTCIceCandidatePair/local}} and {{RTCIceCandidatePair/remote}} attributes of <var>candidatePair</var> do not match a pairing of {{RTCIceCandidatePairEvent/local}} and {{RTCIceCandidatePairEvent/remote}} respectively sent in {{RTCIceTransport/onicecandidatepairadd}}, [= exception/throw =] a {{NotFoundError}}.
+                If |candidatePair| does not [= candidate pair match | match =] any item in [=this=].{{RTCIceTransport/[[CandidatePairs]]}}, [= exception/throw =] a {{NotFoundError}}.
+              </p>
+            </li>
+            <li>
+              <p>
+                [= list/Remove =] all items that [= candidate pair match | match =] |candidatePair| from [=this=].{{RTCIceTransport/[[CandidatePairs]]}}.
               </p>
             </li>
             <li>
--- a/index.html
+++ b/index.html
@@ -897,7 +897,7 @@ dictionary RTCEncodingOptions {
         <ol>
           <li>
             <p>
-              [= list/Remove =] all items that [= candidate pair match | match =] |candidatePair| from |transport|.{{RTCIceTransport/[[CandidatePairs]]}}.
+              [= list/Remove =] |candidatePair| from |transport|.{{RTCIceTransport/[[CandidatePairs]]}}.
             </p>
           </li>
           <li>
--- a/index.html
+++ b/index.html
@@ -1097,7 +1097,8 @@ dictionary RTCEncodingOptions {
             </li>
             <li>
               <p>
-                If |candidatePair| does not [= candidate pair match | match =] any item in [=this=].{{RTCIceTransport/[[CandidatePairs]]}}, [= exception/throw =] a {{NotFoundError}}.
+                If |candidatePair| does not [= candidate pair match | match =] any item in [=this=].
+                {{RTCIceTransport/[[CandidatePairs]]}}, [= exception/throw =] a {{NotFoundError}}.
               </p>
             </li>
             <li>
--- a/index.html
+++ b/index.html
@@ -1007,7 +1007,8 @@ dictionary RTCEncodingOptions {
             </li>
             <li>
               <p>
-                If |candidatePair| does not [= candidate pair match | match =] any item in [=this=].{{RTCIceTransport/[[CandidatePairs]]}}, [= exception/throw =] a {{NotFoundError}}.
+                If |candidatePair| does not [= candidate pair match | match =] any item in [=this=].
+                {{RTCIceTransport/[[CandidatePairs]]}}, [= exception/throw =] a {{NotFoundError}}.
               </p>
             </li>
             <li>
--- a/index.html
+++ b/index.html
@@ -749,6 +749,11 @@ dictionary RTCEncodingOptions {
       </li>
       <li>
         <p>
+          [= list/Append =] |candidatePair| to {{RTCIceTransport/[[CandidatePairs]]}}.
+        </p>
+      </li>
+      <li>
+        <p>
           [= Fire an event =] named
           {{RTCIceTransport/icecandidatepairadd}} at |transport|, using {{RTCIceCandidatePairEvent}},
           with the {{RTCIceCandidatePairEvent/local}} and {{RTCIceCandidatePairEvent/remote}} attributes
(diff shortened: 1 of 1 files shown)


henbos (approved):
LGTM

index.html
```diff
@@ -881,14 +886,29 @@ <h3>
...
+        </ol>
       </li>
     </ol>
     <p>
-      The {{RTCIceTransport}} object is extended by adding the following internal slot:
+      The {{RTCIceTransport}} object is extended by adding the following internal slots:
     </p>
     <ul>
+      <li>
+        <dfn data-dfn-for="RTCIceTransport">[[\CandidatePairs]]</dfn> initialized to an empty list.
```
jan-ivar, line 910: Normally I'd say use an [ordered set](https://infra.spec.whatwg.org/#ordered-set) instead, but I guess a list is fine since implementations can't rely on object match anyway since pairs are dictionaries with a match algorithm.

This means it's up to algorithms to keep out duplicates.

index.html
```diff
@@ -750,6 +750,11 @@ <h3>
           Let |transport:RTCIceTransport| be the {{RTCIceTransport}} object associated with |candidatePair|.
         </p>
       </li>
+      <li>
+        <p>
+          [= list/Append =] |candidatePair| to {{RTCIceTransport/[[CandidatePairs]]}}.
+        </p>
+      </li>
```
jan-ivar, line 757: Github won't let me comment on the line directly above this section, which is:
> Let |candidatePair:RTCIceCandidatePair| be the candidate pair that was [= formed =].

While readers can hover over the `candidatePair` variable to learn its type, I think we need to spell out in prose what the type is now, since to a non-interactive reader `candidatePair` reads like an internal concept so far in this algorithm. We need to expound on its type before adding it to [[CandidatePairs]] since objects there need to be usable with the new match algorithm.

Something like:

> Let |candidatePair:RTCIceCandidatePair| be a new `RTCIceCandidate` dictionary representing the candidate pair that was [= formed =].

Moreover, we probably also need to clarify where the two instances of RTCIceCandidate interface objects are coming from. Are they also created here using `new`?

index.html
```diff
@@ -881,14 +886,29 @@ <h3>
       </li>
       <li>
         <p>
-          Otherwise, instruct the [= ICE agent =] to remove the candidate pair indicated by |candidatePair|.
+          Otherwise (if |accepted| is <code>true</code>), run the following steps:
         </p>
+        <ol>
+          <li>
+            <p>
+              [= list/Remove =] all items that [= candidate pair match | match =] |candidatePair| from |transport|.{{RTCIceTransport/[[CandidatePairs]]}}.
```
jan-ivar, line 894: Same here, I'd prefer identifying the candidatePair in the internal slot higher up in this algorithm where it says: 
```html
  <p>
    Let |candidatePair:RTCIceCandidatePair| be the candidate pair which is being removed.
  </p>
```
to
```html
  <p>
    [=Assert=]: the candidate pair which is being removed [=candidate pair match|matches=]
    exactly one item in |transport|.{{RTCIceTransport/[[CandidatePairs]]}}.
  </p>
</li>
<li>
  <p>
    Let |candidatePair:RTCIceCandidatePair| be the item in
    |transport|.{{RTCIceTransport/[[CandidatePairs]]}} which is being removed.
 </p>
```
After those changes, this line becomes:
```suggestion
              [= list/Remove =] |candidatePair| from |transport|.{{RTCIceTransport/[[CandidatePairs]]}}.
```

index.html
```diff
@@ -981,8 +1001,7 @@ <h2>Methods</h2>
             </li>
             <li>
               <p>
-                If |candidatePair| does not describe a candidate pair formed for [=this=] {{RTCIceTransport}} and sent in
-                {{RTCIceTransport/onicecandidatepairadd}}, [= exception/throw =] a {{NotFoundError}}.
+                If |candidatePair| does not [= candidate pair match | match =] any item in [=this=].{{RTCIceTransport/[[CandidatePairs]]}}, [= exception/throw =] a {{NotFoundError}}.
```
jan-ivar, line 1004: Please wordwrap.

index.html
```diff
@@ -1071,7 +1090,12 @@ <h2>Methods</h2>
             </li>
             <li>
               <p>
-                If the {{RTCIceCandidatePair/local}} and {{RTCIceCandidatePair/remote}} attributes of <var>candidatePair</var> do not match a pairing of {{RTCIceCandidatePairEvent/local}} and {{RTCIceCandidatePairEvent/remote}} respectively sent in {{RTCIceTransport/onicecandidatepairadd}}, [= exception/throw =] a {{NotFoundError}}.
+                If |candidatePair| does not [= candidate pair match | match =] any item in [=this=].{{RTCIceTransport/[[CandidatePairs]]}}, [= exception/throw =] a {{NotFoundError}}.
```
jan-ivar, line 1093: Please wordwrap

index.html
```diff
@@ -1071,7 +1090,12 @@ <h2>Methods</h2>
...
+              </p>
+            </li>
+            <li>
+              <p>
+                [= list/Remove =] all items that [= candidate pair match | match =] |candidatePair| from [=this=].{{RTCIceTransport/[[CandidatePairs]]}}.
```
jan-ivar, line 1098: Now we can eliminate "all" here. E.g.
```suggestion
                [= list/Remove =] the item in
                [=this=].{{RTCIceTransport/[[CandidatePairs]]}} that
                [= candidate pair match | matches =] |candidatePair|.
```

jan-ivar (changes requested):
Thanks for the icleanup. I'd like to specify some more invariants around this internal slot.

index.html
```diff
@@ -1071,7 +1090,12 @@ <h2>Methods</h2>
...
               </p>
             </li>
             <li>
```
jan-ivar, line 1101: Also, we probably need to clear [[CandidatePairs]] on ICE restart.

index.html
alvestrand: The nice thing about "all" is that it implicitly addresses the case where there are no matching items.
We can append ", if any" to the suggestion to keep that effect.

index.html
```diff
@@ -1071,7 +1097,12 @@ <h2>Methods</h2>
...
               <p>
-                If the {{RTCIceCandidatePair/local}} and {{RTCIceCandidatePair/remote}} attributes of <var>candidatePair</var> do not match a pairing of {{RTCIceCandidatePairEvent/local}} and {{RTCIceCandidatePairEvent/remote}} respectively sent in {{RTCIceTransport/onicecandidatepairadd}}, [= exception/throw =] a {{NotFoundError}}.
+                If |candidatePair| does not [= candidate pair match | match =] any item in [=this=].{{RTCIceTransport/[[CandidatePairs]]}}, [= exception/throw =] a {{NotFoundError}}.
```
henbos, line 1100:
```suggestion
                If |candidatePair| does not [= candidate pair match | match =] any item in [=this=].
                {{RTCIceTransport/[[CandidatePairs]]}}, [= exception/throw =] a {{NotFoundError}}.
```

index.html
henbos: Good question, I filed #196 to decide if we need to add more steps or if this is already covered by the existing removal steps, including whether or not the `onicecandidatepairremoved` should fire?

alvestrand:
Merged as #197

Issue #217, opened 2024-08-06
ReSpec errors from duplicate definitions of RTCIceCandidatePair


sam-vi:
Ref: https://github.com/w3c/webrtc-pc/issues/2930.

With the conversion of `RTCIceCandidatePair` from a dictionary to an interface, the XRef database contains some ambiguous duplicate definitions with differing types from multiple version of the "webrtc" spec with the same spec ID:
- `RTCIceCandidatePair`: https://respec.org/xref/?term=RTCIceCandidatePair&specs=webrtc&types=_IDL_
- `RTCIceCandidatePair.local`: https://respec.org/xref/?term=local&specs=webrtc&types=attribute%2Cconst%2Cdict-member&for=RTCIceCandidatePair
- `RTCIceCandidatePair.remote`: https://respec.org/xref/?term=remote&specs=webrtc&types=attribute%2Cdict-member%2Cconst&for=RTCIceCandidatePair

This leads to validation errors in [webrtc-extensions](https://w3c.github.io/webrtc-extensions/#respec-offender-definition-is-ambiguous) when linking to these definitions.

Possibly related bug: speced/respec#4307.

dontcallmedom:
the "right" way to fix this will involve re-publishing the current editors draft of the WebRTC spec on /TR, which mostly needs approval from the WG

dontcallmedom:
With the republication of the WebRTC spec on TR, that problem is now resolved

PR #221, opened 2024-08-29
Add scaleResolutionDownTo to RTCRtpEncodingParameters

### LINKED ISSUE #159: RTCRtpEncodingParameters: scaleResolutionDownTo
We all know and love `scaleResolutionDownBy`... it let's you do this:

Capture 720p and apply expensive video effects on the track.
Send {active, scaleResolutionDownBy:1 (720p)} + {active, scaleResolutionDownBy:2 (360p)} simulcast.

But then server tells us 720p is not needed, so we send {inactive, active:360p}.
Now you might wonder, why are we applying expensive video effects on a 720p track if we're only sending it in 360p?

So we track.applyConstraints() to 1/2 downscale the track and we setParameters() to 2x the scaling factors to counter the fact that the input frame size is smaller. The end result is:

360p track sent as {inactive} + {active, scaleResolutionDownBy:1 (still 360p)}.

**Problem:** This is racy. The track changing size and the parameters updating the scaling factors are not in-sync, so you might send 720p on the VGA layers for a few frames. You're wasting a key frame on sending the wrong resolution and then you're wasting another key frame to adjust back to the desired 360p resolution. Working around this by temporarily inactivating all encodings while the track is resizing will appear like glitches to the receiver of the stream and there probably would still have to be key frames when the encodings are re-activated. The API was not designed for this.

**Proposal:** Add "scaleResolutionTo: 360" API which is equivalent to always having the correct scaling factor. In the above example, changing the track's size from 720p to 360p would not result in reconfiguring the encoder and there would be no key frames or races.

henbos:
Fixes #159.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-extensions/pull/221.html" title="Last updated on Sep 12, 2024, 3:34 PM UTC (0ec01b0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-extensions/221/5aaf7b2...henbos:0ec01b0.html" title="Last updated on Sep 12, 2024, 3:34 PM UTC (0ec01b0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -375,8 +375,8 @@ partial interface RTCRtpTransceiver {
           applied when this parameter is used. Instead, frames are either sent
           as-is or they are downscaled by the encoder to uphold the
           restrictions. The encoder will never upscale a frame. In simulcast,
-          this can result in dropping top layer(s) if the input frame is too
-          small as to avoid sending the same resolution on multiple layers.</p>
+          encodings are sent up to and including the layer that results in a
+          downscale, ignoring any layers further up.</p>
           <p>When setting parameters, if {{resolutionRestriction}} is specified
           on any encoding, check that the following is true or else throw an
           {{InvalidModificationError}}:</p>
--- a/index.html
+++ b/index.html
@@ -461,7 +461,7 @@ partial interface RTCRtpTransceiver {
         <dd>
           <p>The maximum height that frames will be encoded with.
           The restrictions are orientation agnostic, see note below.
-          When scaling is applied, both dimensions of the frame are downscaled
+          When scaling is applied, both dimensions of the frame MUST be downscaled
           using the same factor.</p>
         </dd>
         <p class="note">
--- a/index.html
+++ b/index.html
@@ -372,7 +372,7 @@ partial interface RTCRtpTransceiver {
         <dd>
           <p>The maximum dimensions at which to restrict this encoding.</p>
           <p>When {{scaleResolutionDownTo}} is specified, the
-          {{RTCRtpEncodingParameters/scaleResolutionDownBy}} scaling factor MUST
+          {{RTCRtpEncodingParameters/scaleResolutionDownBy}} value MUST
           be ignored by the encoder. Instead, frames are sent according to the
           specified resolution restrictions: frames MUST either be downscaled or
           sent as-is. The encoder MUST NOT upscale a frame. In simulcast, not
--- a/index.html
+++ b/index.html
@@ -459,8 +459,8 @@ partial interface RTCRtpTransceiver {
           <dfn data-idl>maxHeight</dfn> of type <span class="idlMemberType">unsigned long</span>
         </dt>
         <dd>
-          <p>The maximum height that frames will be encoded with without getting
-          downscaled. The restrictions are orientation agnostic, see note below.
+          <p>The maximum height that frames will be encoded with.
+          The restrictions are orientation agnostic, see note below.
           When scaling is applied, both dimensions of the frame are downscaled
           using the same factor.</p>
         </dd>
--- a/index.html
+++ b/index.html
@@ -359,7 +359,7 @@ partial interface RTCRtpTransceiver {
       additional members to control audio packetization.
     </p>
     <pre class="idl">partial dictionary RTCRtpEncodingParameters {
-    RTCResolutionRestriction restrictedResolution;
+    RTCResolutionRestriction resolutionRestriction;
     unsigned long            ptime;
     boolean                  adaptivePtime = false;
 };</pre>
@@ -367,7 +367,7 @@ partial interface RTCRtpTransceiver {
       <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
       <dl data-link-for="RTCRtpEncodingParameters" data-dfn-for="RTCRtpEncodingParameters" class="dictionary-members">
         <dt>
-          <dfn data-idl>restrictedResolution</dfn> of type <span class="idlMemberType">{{RTCResolutionRestriction}}</span>
+          <dfn data-idl>resolutionRestriction</dfn> of type <span class="idlMemberType">{{RTCResolutionRestriction}}</span>
         </dt>
         <dd>
           <p>The resolution at which to restrict this encoding.</p>
@@ -377,12 +377,12 @@ partial interface RTCRtpTransceiver {
           restrictions. The encoder will never upscale a frame. In simulcast,
           this can result in dropping top layer(s) if the input frame is too
           small as to avoid sending the same resolution on multiple layers.</p>
-          <p>When setting parameters, if {{restrictedResolution}} is specified
+          <p>When setting parameters, if {{resolutionRestriction}} is specified
           on any encoding, check that the following is true or else throw an
           {{InvalidModificationError}}:</p>
           <ul>
             <li>
-              <p>{{restrictedResolution}} is specified on all encodings and has
+              <p>{{resolutionRestriction}} is specified on all encodings and has
               values in both dimensions that are greater than 0.</p>
             </li>
             <li>
@@ -464,7 +464,7 @@ partial interface RTCRtpTransceiver {
         in landscape mode but the frame is in portrait mode or vice versa, the
         encoder will internally swap the restricted width and height prior to
         each size comparison. This does not modify the value of
-        {{RTCRtpEncodingParameters/restrictedResolution}}.
+        {{RTCRtpEncodingParameters/resolutionRestriction}}.
       </p>
     </section>
   </section>
--- a/index.html
+++ b/index.html
@@ -372,15 +372,9 @@ partial interface RTCRtpTransceiver {
         <dd>
           <p>The maximum dimensions at which to restrict this encoding.</p>
           <p>When {{scaleResolutionDownTo}} is specified, the
-          {{RTCRtpEncodingParameters/scaleResolutionDownBy}} value MUST
-          be ignored. Instead, frames are sent according to the
-          specified resolution restrictions: frames MUST NOT be upscaled. In simulcast, not
-          upscaling can result in multiple same-sized encodings if the size of
-          the frame is smaller than {{scaleResolutionDownTo}}; an encoding MUST
-          NOT be sent if {{scaleResolutionDownTo}} is larger than the frame and
-          there exists another {{RTCRtpEncodingParameters/active}} encoding with
-          a smaller {{scaleResolutionDownTo}} density resulting in the same
-          size.</p>
+          {{RTCRtpEncodingParameters/scaleResolutionDownBy}} value MUST be
+          ignored. Instead, frames are sent according to the specified
+          resolution restrictions: frames MUST NOT be upscaled.</p>
           <p>When configuring parameters, the following validation MUST be
           performed if {{scaleResolutionDownTo}} is specified on any encoding or
           else {{RTCPeerConnection/addTransceiver()}} [= exception/throws =] a
@@ -450,19 +444,19 @@ partial interface RTCRtpTransceiver {
           <dfn data-idl>maxWidth</dfn> of type <span class="idlMemberType">unsigned long</span>
         </dt>
         <dd>
-          <p>The maximum width that frames will be encoded with.
-          The restrictions are orientation agnostic, see note below.
-          When scaling is applied, both dimensions of the frame are downscaled
-          using the same factor.</p>
+          <p>The maximum width that frames will be encoded with. The
+          restrictions are orientation agnostic, see note below. When scaling is
+          applied, both dimensions of the frame are downscaled using the same
+          factor.</p>
         </dd>
         <dt>
           <dfn data-idl>maxHeight</dfn> of type <span class="idlMemberType">unsigned long</span>
         </dt>
         <dd>
-          <p>The maximum height that frames will be encoded with.
-          The restrictions are orientation agnostic, see note below.
-          When scaling is applied, both dimensions of the frame MUST be downscaled
-          using the same factor.</p>
+          <p>The maximum height that frames will be encoded with. The
+          restrictions are orientation agnostic, see note below. When scaling is
+          applied, both dimensions of the frame MUST be downscaled using the
+          same factor.</p>
         </dd>
         <p class="note">
           The restrictions being orientation agnostic means that they will
--- a/index.html
+++ b/index.html
@@ -449,17 +449,26 @@ partial interface RTCRtpTransceiver {
         </dt>
         <dd>
           <p>The maximum width that frames will be encoded with without getting
-          downscaled. When scaling is applied, both dimensions of the frame are
-          downscaled using the same factor.</p>
+          downscaled. The restrictions are orientation agnostic, see note below.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -359,12 +359,38 @@ <h3>
       additional members to control audio packetization.
     </p>
     <pre class="idl">partial dictionary RTCRtpEncodingParameters {
-  unsigned long ptime;
-  boolean adaptivePtime = false;
+    RTCResolutionRestriction restrictedResolution;
```
alvestrand, line 362: bikeshed: resolutionRestriction?
The parameter isn't a resolution, it's a restriction on resolution.

index.html
```diff
@@ -359,12 +359,38 @@ <h3>
...
+        <dt>
+          <dfn data-idl>restrictedResolution</dfn> of type <span class="idlMemberType">{{RTCResolutionRestriction}}</span>
+        </dt>
+        <dd>
+          <p>The resolution at which to restrict this encoding.</p>
+          <p>The default "scale resolution down by" factors (e.g. 4:2:1) are not
+          applied when this parameter is used. Instead, frames are either sent
+          as-is or they are downscaled by the encoder to uphold the
+          restrictions. The encoder will never upscale a frame. In simulcast,
+          this can result in dropping top layer(s) if the input frame is too
```
alvestrand, line 378: Define "top layer". We should either send the first encoding in the list for which the restriction results in an unscaled frame or the last encoding in the list for which the restriction results in an unscaled frame, and not send the frame on the rest of the encodings.

index.html
```diff
@@ -403,6 +429,45 @@ <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
...
+};</pre>
+    <section id="rtcresolutionrestriction-attributes">
+      <h2>Dictionary {{RTCResolutionRestriction}} Members</h2>
+      <dl data-link-for="RTCResolutionRestriction" data-dfn-for="RTCResolutionRestriction" class="dictionary-members">
+        <dt>
+          <dfn data-idl>maxWidth</dfn> of type <span class="idlMemberType">unsigned long</span>
+        </dt>
+        <dd>
+          <p>The maximum width that frames will be encoded with without getting
+          downscaled. The size comparison is orientation agnostic. When scaling
```
alvestrand, line 448: Define "orientation agnostic" - either that X is always X no matter what the CVO rotation parameter says, or that if width > height, width is X and height is Y, or vice versa.

alvestrand: Defined further down, it's the swap meaning that's intended. I worry about this in a world of tons of vertical video.

henbos: In a world of [VideoFrame](https://w3c.github.io/webcodecs/#videoframe) and effects processing, I think this just needs to work when we inspect the frame WxH. Or is there a different way to do this that is better?

index.html
henbos: Makes sense, done

index.html
henbos: Updated this sentence, is this better?

index.html
henbos: The intent is that you shouldn't need to call setParamters() every time the orientation changes in order to get the same scaling factors

henbos: If I express this in terms of [primary orientation](https://w3c.github.io/mediacapture-main/#dfn-primary-orientation) instead, then specifying 1280x720 should work regardless of orientation, but it would in addition also allow me to enter a 720x1280 and force it to fit into that? Is that correct + what we should do?

henbos: Feedback from meeting: the UA should calculate pixel count from W and H and use that as the restriction

henbos: Also: Jan-Ivar liked the name scaleResolutionDownTo better, will update

henbos: Calculating pixel count internally as an implementation for "orientation agnostic" does not work when the restriction is not the same aspect ratio as the frame.

Example: frame is 1280x720 and I want to restrict it to 360 in height, but I care less about width and say {maxWidth:2000, maxHeight:360}. What should happen to get 360p and maintain aspect ratio is a 640x360 frame, but if we internally calculate a really large pixel count (2000*360) I will end up with a frame that is much taller than 360.

I think we need to either swap W and H on the fly (current PR), or not be orientation agnostic, or swap W and H but do so based on orientation metadata rather than trying to infer it from input frame's dimensions.

henbos: As discussed, I removed the talk about orientation since the rotation is separate from what is width and height meaning the frame is 1280x720 regardless of orientation and we can always say width and height without any tweaks to this.

I also changed the name back to scaleResolutionDownTo and updated the text about when not to send duplicate encodings.

index.html
```diff
@@ -403,6 +433,46 @@ <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
...
+          When scaling is applied, both dimensions of the frame are downscaled
+          using the same factor.</p>
+        </dd>
+        <p class="note">
+          The restrictions being orientation agnostic means that they will
+          automatically be adjusted to the orientation of the frame being
+          restricted (portrait mode or landscape mode) by swapping width and
+          height if necessary. This means that it does not matter if 1280x720 or
+          720x1280 is specified, both always result in the exact same scaling
+          factor regardless of the orientation of the frame.
```
henbos, line 471: I confirmed that we need to do this and that the C++ implementation already does this.

index.html
```diff
@@ -359,12 +359,42 @@ <h3>
...
+          <p>The default "scale resolution down by" factors (e.g. 4:2:1) are not
+          applied when this parameter is used. Instead, frames are either sent
+          as-is or they are downscaled by the encoder to uphold the
+          restrictions. The encoder will never upscale a frame. In simulcast,
+          not upscaling can result in multiple same-sized encodings if the size
+          of the frame is smaller than {{scaleResolutionDownTo}}; an encoding is
+          not sent if {{scaleResolutionDownTo}} is larger than the frame and
+          there exists another {{RTCRtpEncodingParameters/active}} encoding with
+          a smaller {{scaleResolutionDownTo}} density resulting in the same
+          size.</p>
```
alvestrand, line 383: Does this mean that if we have a 480x360 stream, and there are two encodings:
scaleResolutionDownTo(480x360)
scaleResolutionDownTo(1024x768)
the first one (the "smaller density") is the one that will be sent, no matter which order they occur in?
This makes intuitive sense to me. Be careful of the case where the densities are equal - in that case, the current language says that both will be sent.

henbos: Correct! I figured if scaleBy allows any order then scaleTo should too. In practice the libwebrtc bitrate allocation logic probably assumes smaller-to-larger but I'm not aware of anything in the spec that says you have to use this order, so I made this "don't send" logic agnostic towards order too

index.html
```diff
@@ -359,12 +359,43 @@ <h3>
...
+          <p>When setting parameters, if {{scaleResolutionDownTo}} is specified
```
jan-ivar, line 384: Maybe add a note this is both for addTransceiver and setParameters

index.html
```diff
@@ -359,12 +359,43 @@ <h3>
...
 };</pre>
     <section id="rtcrtpencodingparameters-attributes">
       <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
       <dl data-link-for="RTCRtpEncodingParameters" data-dfn-for="RTCRtpEncodingParameters" class="dictionary-members">
+        <dt>
+          <dfn data-idl>scaleResolutionDownTo</dfn> of type <span class="idlMemberType">{{RTCResolutionRestriction}}</span>
+        </dt>
+        <dd>
+          <p>The maximum dimensions at which to restrict this encoding.</p>
+          <p>The default "scale resolution down by" factors (e.g. 4:2:1) are not
```
jan-ivar, line 374: please rephrase without air quotes

index.html
```diff
@@ -359,12 +359,43 @@ <h3>
...
+          applied when this parameter is used. Instead, frames are either sent
+          as-is or they are downscaled by the encoder to uphold the
+          restrictions. The encoder will never upscale a frame. In simulcast,
```
jan-ivar, line 377:
```suggestion
          restrictions. The encoder MUST NOT upscale a frame. In simulcast,
```

index.html
```diff
@@ -359,12 +359,43 @@ <h3>
...
+          promise [= rejected =] with a newly [= exception/created =]
+          {{InvalidModificationError}}:</p>
+          <ul>
+            <li>
+              <p>{{scaleResolutionDownTo}} is specified on all encodings and has
+              values in both dimensions that are greater than 0.</p>
+            </li>
+            <li>
+              <p>{{RTCRtpEncodingParameters/scaleResolutionDownBy}} is not
+              specified on any encoding.</p>
```
henbos, line 395: Instead of throwing when this is specified, say to ignore it. This is the way to make get+setParameters compatible with the default factors which today are different on different browsers (due to bugs), i.e. undefined on Chrome, 4:2:1 on Firefox and 1:1:1 on Safari

index.html
henbos: As per other comment we can remove this sentence because we'll say to ignore it later in the text

jan-ivar: The issue with the existing `scaleResolutionDownBy` values is that they effectively have default values (inserted in prose even though it's not in WebIDL since the defaults are dynamic, e.g. 4,2,1.

Unfortunately, webcompat here is poor since pasting this into web console produces 4,2,1 values in Firefox, 1,1,1, in Safari, and undefined, undefined, undefined in Chrome as defaults for `scaleResolutionDownBy`: 
```js
new RTCPeerConnection().addTransceiver("video", {sendEncodings: [{rid: "a"}, {rid: "b"}]}).sender.getParameters().encodings;
```

Hopefully this means few websites are actually relying on defaults or they're going to have a hard time.

Long-term we should fix the defaults to be better, but in the short-term we cannot rely on them. Better to ignore them I think

jan-ivar: This is [crbug 344943229](https://issues.chromium.org/issues/344943229) in Chrome and [webkit 279202](https://bugs.webkit.org/show_bug.cgi?id=279202) in Safari.

henbos: Done

index.html
henbos: Done. Also added MUSTs and MUST NOTs in other places as appropriate.

index.html
henbos: Done

index.html
henbos: Done

henbos:
The "Validate and Auto Publish" steps are failing due to #217, i.e. there are no issues in this PR

index.html
```diff
@@ -359,12 +359,45 @@ <h3>
...
     <section id="rtcrtpencodingparameters-attributes">
       <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
       <dl data-link-for="RTCRtpEncodingParameters" data-dfn-for="RTCRtpEncodingParameters" class="dictionary-members">
+        <dt>
+          <dfn data-idl>scaleResolutionDownTo</dfn> of type <span class="idlMemberType">{{RTCResolutionRestriction}}</span>
+        </dt>
+        <dd>
+          <p>The maximum dimensions at which to restrict this encoding.</p>
+          <p>When {{scaleResolutionDownTo}} is specified, the
+          {{RTCRtpEncodingParameters/scaleResolutionDownBy}} scaling factor MUST
```
jan-ivar, line 375:
```suggestion
          {{RTCRtpEncodingParameters/scaleResolutionDownBy}} value MUST
```

index.html
```diff
@@ -359,12 +359,45 @@ <h3>
...
+          be ignored by the encoder. Instead, frames are sent according to the
```
jan-ivar, line 376:
```suggestion
          be ignored. Instead, frames are sent according to the
```

index.html
```diff
@@ -359,12 +359,45 @@ <h3>
...
+          specified resolution restrictions: frames MUST either be downscaled or
+          sent as-is. The encoder MUST NOT upscale a frame. In simulcast, not
```
jan-ivar, line 378:
```suggestion
          specified resolution restrictions: frames MUST NOT be upscaled. In simulcast, not
```

index.html
```diff
@@ -403,6 +436,46 @@ <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
...
+};</pre>
+    <section id="rtcresolutionrestriction-attributes">
+      <h2>Dictionary {{RTCResolutionRestriction}} Members</h2>
+      <dl data-link-for="RTCResolutionRestriction" data-dfn-for="RTCResolutionRestriction" class="dictionary-members">
+        <dt>
+          <dfn data-idl>maxWidth</dfn> of type <span class="idlMemberType">unsigned long</span>
+        </dt>
+        <dd>
+          <p>The maximum width that frames will be encoded with without getting
+          downscaled. The restrictions are orientation agnostic, see note below.
```
jan-ivar, line 455:
```suggestion
          <p>The maximum width that frames will be encoded with.
          The restrictions are orientation agnostic, see note below.
```

index.html
```diff
@@ -403,6 +436,46 @@ <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
...
+          When scaling is applied, both dimensions of the frame are downscaled
+          using the same factor.</p>
+        </dd>
+        <dt>
+          <dfn data-idl>maxHeight</dfn> of type <span class="idlMemberType">unsigned long</span>
+        </dt>
+        <dd>
+          <p>The maximum height that frames will be encoded with without getting
+          downscaled. The restrictions are orientation agnostic, see note below.
```
jan-ivar, line 464:
```suggestion
          <p>The maximum height that frames will be encoded with.
          The restrictions are orientation agnostic, see note below.
```

index.html
```diff
@@ -403,6 +436,46 @@ <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
...
+          When scaling is applied, both dimensions of the frame are downscaled
+          using the same factor.</p>
```
jan-ivar, line 465:
```suggestion
          When scaling is applied, both dimensions of the frame MUST be downscaled
          using the same factor.</p>
```

index.html
```diff
@@ -359,12 +359,44 @@ <h3>
...
+          <p>The maximum dimensions at which to restrict this encoding.</p>
+          <p>When {{scaleResolutionDownTo}} is specified, the
+          {{RTCRtpEncodingParameters/scaleResolutionDownBy}} value MUST
+          be ignored. Instead, frames are sent according to the
+          specified resolution restrictions: frames MUST NOT be upscaled. In simulcast, not
+          upscaling can result in multiple same-sized encodings if the size of
+          the frame is smaller than {{scaleResolutionDownTo}}; an encoding MUST
+          NOT be sent if {{scaleResolutionDownTo}} is larger than the frame and
+          there exists another {{RTCRtpEncodingParameters/active}} encoding with
+          a smaller {{scaleResolutionDownTo}} density resulting in the same
```
jan-ivar, line 382: Can we defer this part about simulcast to another issue? This is supposed to be a max value which implies lower resolutions should be sent, but here it says the opposite. Might need a bit more discussion

henbos: Yes, I filed #222

jan-ivar (approved):
approve with simulcast mention removed

PR #243, opened 2025-09-26
Add onssrcchange/oncsrcchange events to RTCRtpReceiver

### LINKED ISSUE #230: Feature request: Add event for CSRC change
Something that would be very useful to myself and others would be an event for when CSRCs change.

Currently, we are forced to poll getContributingSources which can be a bit cumbersome. This would widely signal that media is flowing

henbos:
Fixes #230.

To remind everyone, this was presented to the WG back in November of last year with resolution "RESOLUTION: Rough consensus on moving forward with events for csrc/ssrc changes, with concerns noted on related onunmute interop".
- WG notes: https://www.w3.org/2024/11/19-webrtc-minutes.html#c7d9
- Video recording: https://www.youtube.com/watch?v=eHjg9figxLk&t=1h11m50s

Regarding the onunmute discussion, Chrome has attempted to ship "remote track onunmute in response to RTP" but that had to be rolled back due to bugs detected at Stable, hopefully we can roll that forward again. I still believe that issue is largely tangential to this issue.

@koumiGoogle @taylor-b @jan-ivar

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-extensions/pull/243.html" title="Last updated on Oct 27, 2025, 9:38 AM UTC (5f6070c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-extensions/243/ccc0614...henbos:5f6070c.html" title="Last updated on Oct 27, 2025, 9:38 AM UTC (5f6070c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1468,36 +1468,46 @@ partial interface RTCRtpTransceiver {
   attribute EventHandler oncsrcchange;
 };</pre>
     <p>
-      Whenever the {{RTCRtpContributingSource/source}} identifier is updated based on the last decoded frame,
-      the [= user agent =] MUST [= queue a task =] to [= fire an event =] named {{RTCRtpReceiver/ssrcchange}}
-      if the SSRC changed and an event named {{RTCRtpReceiver/csrcchange}} if the CSRC changed. If both
-      identifiers are changed based on the same decoded frame, both events fire in the same queued task. This
-      ensures that both SSRC and CSRC information is exposed to the app at the same time.
+      Whenever the {{RTCRtpContributingSource/source}} identifier(s) are updated based on the last decoded
+      frame as defined <a data-cite="WEBRTC#dom-rtcrtpcontributingsource">here</a>, the [= user agent =]
+      MUST append the following steps to the end of the same [=queue a task|queued task=]:
     </p>
-      <h3>
-        Attributes
-      </h3>
-      <dl data-link-for="RTCRtpReceiver" data-dfn-for="RTCRtpReceiver" class="attributes">
-        <dt>
-          <dfn id="dom-rtptransceiver-onssrcchange">onssrcchange</dfn> of type {{EventHandler}}
-        </dt>
-        <dd>
-          <p>
-            Fired when the {{RTCRtpContributingSource/source}} in {{RTCRtpReceiver/getSynchronizationSources()}}
-            changes, see above.
-          </p>
-        </dd>
-        <dt>
-          <dfn id="dom-rtptransceiver-oncsrcchange">oncsrcchange</dfn> of type {{EventHandler}}
-        </dt>
-        <dd>
-          <p>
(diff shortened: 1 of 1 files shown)


henbos:
I believe that receiver.onssrcchange can NOT be used to shim track.onunmute because onssrcchange fires at decode time while onunmute fires at packet reception time, and in Chrome we don't decode unless we have a sink

jan-ivar:
> ... concerns noted on related onunmute interop".

My preference would be to resolve https://github.com/w3c/webrtc-pc/issues/3077 first before merging this.

henbos:
> My preference would be to resolve https://github.com/w3c/webrtc-pc/issues/3077 first before merging this.

I intend to present both at then next WG meeting and it sounds like we have agreement there.

But to clarify this comment I made:

>  I believe that receiver.onssrcchange can NOT be used to shim track.onunmute because onssrcchange fires at decode time while onunmute fires at packet reception time, and in Chrome we don't decode (audio) unless we have a sink

What I mean is that one of the use cases for unmute that has been used in examples is `track.onunmute = () => mediaTag.srcObject = stream /w track;` and for audio trying to do this: `track.onssrcchange = () => audioTag.srcObject = stream /w track;` would simply never fire because you need to set srcObject first in order to unblock decoding and you need decoding for onssrcchange to fire. But for `track.onunmute` you don't need to have decoding happening first because it fires as soon as the packet is received, which is also a good think for the "set srcObject" use case because attaching sinks at packet reception means you have &lt;jitter buffer length&gt; amount of time to set the srcObject without dropping any frame. If you were reacting to "frame decoded" then that frame would likely be dropped because you didn't set srcObject in time. So it's an argument for "we need both APIs regardless".

jan-ivar:
My understanding is SSRC lives in the RTP header, available as soon as an SRTP-authenticated packet arrives, long before any depacketization or codec decode.

> audioTag.srcObject = stream /w track; would simply never fire because you need to set srcObject first in order to unblock decoding and you need decoding for onssrcchange to fire.

Chrome is skipping more than decode absent a sink if this is true. This seems like an implementation optimization, and not something we want to standardize.

I can't think of a reason other than implementation convenience to limit the API to when sinks exist. Conceptually, I'd expect the event to fire regardless of sink.

henbos:
And we don’t want to disallow UAs from optimizing

jan-ivar:
> Sounds like we agree that "optimize versus accuracy" is in the UA territory meaning we cannot shim one of the APIs over the other which means onssrcchange is independent from onunmute

I don't think we can arrive at that. Both [unmute](https://w3c.github.io/webrtc-pc/#mediastreamtrack-network-use) and `ssrcchange` seem defined entirely by the RTP stack. So any implementation skipping that over a missing sink does not appear to be compliant. Personally, I also question what this would be optimizing for.

PR #252, opened 2026-05-08
Encoding complexity

### LINKED ISSUE #191: Add API to control encode complexity
## Background

Encode complexity settings are hardcoded for WebRTC's built-in encoders (libvpx VP8/VP9, libaom AV1 and OpenH264). The settings depend on platform, number of CPU cores and video resolution and are optimized to provide acceptable performance on a wide range of devices. In some scenarios these default settings are suboptimal. Access to encode complexity settings would allow applications to optimize the trade-off between device resource usage and compression efficiency for their use cases. For example, a higher encode complexity mode can be used to achieve better video quality and/or to reduce video bitrate.

## Proposed API

Add encodeComplexityMode to [RTCRtpEncodingParameters](https://www.w3.org/TR/webrtc/#dom-rtcrtpencodingparameters):

```
enum RTCEncodeComplexityMode {
  "low",
  "normal",
  "high"
};

partial dictionary RTCRtpEncodingParameters {
  RTCEncodeComplexityMode encodeComplexityMode = "normal";
};
```

**encodeComplexityMode** specifies the encoding complexity mode. "normal" is the default mode that provides a balance between device resource usage and compression efficiency suitable for most use cases. Relative to "normal" mode:
* "low" mode results in lower device resource usage and worse compression efficiency
* "high" mode results in higher device resource usage and better compression efficiency

The user agent SHOULD configure the encoder according to the encoding complexity mode specified. Changes in encoding performance are codec specific and are not guaranteed.

## Details

A hardcoded mapping will be used to convert complexity mode to encoding settings ([CPUUSED](https://source.chromium.org/chromium/chromium/src/+/main:third_party/libaom/source/libaom/aom/aomcx.h;drc=116d912e0ce99de575548002cbe964e3e6381097;l=220) in the case of SW libaom/libvpx encoders, [KEY_COMPLEXITY](https://developer.android.com/reference/android/media/MediaFormat#KEY_COMPLEXITY) in the case of Android MediaCodec, etc). 

Relative differences in encoding performance between different encode complexity modes are not fixed and may change in the new binary due to changes in underlying encoders and/or compilation settings.

sprangerik:
This introduces a new enum `RTCEncodingComplexity` and adds an `encodingComplexity` member to `RTCRtpEncodingParameters` to allow applications to hint at the desired trade-off between quality and CPU/power consumption for an encoding.

Fixes #191

***
<a href="https://pr-preview.s3.amazonaws.com/sprangerik/webrtc-extensions/pull/252.html" title="Last updated on May 8, 2026, 2:38 PM UTC (a505af5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-extensions/252/9f2a0a1...sprangerik:a505af5.html" title="Last updated on May 8, 2026, 2:38 PM UTC (a505af5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -390,13 +390,26 @@ partial interface RTCRtpTransceiver {
           </ol>
         </dd>
       </dl>
-    </section>
-  </section>
-  <section id="rtcrtpencodingparameters">
-    <h3>
-      RTCRtpEncodingParameters extensions
-    </h3>
-    <p>
+    \
+        </section>
+      </section>
+      <section id="rtcencodingcomplexity">
+        <h3>
+          The {{RTCEncodingComplexity}} enum.
+        </h3>
+        <pre class="idl">enum RTCEncodingComplexity {
+        "minimum",
+        "low",
+        "normal",
+        "high",
+        "very-high",
+        "maximum"
+    };</pre>
+      </section>
+      <section id="rtcrtpencodingparameters">
+        <h3>
+          RTCRtpEncodingParameters extensions
+        </h3>
       The {{RTCRtpEncodingParameters}} dictionary is defined in
       [[WEBRTC]]. This document extends that dictionary with
       additional members to control audio packetization.
@@ -405,6 +418,7 @@ partial interface RTCRtpTransceiver {
     RTCResolutionRestriction scaleResolutionDownTo;
     unsigned long            ptime;
     boolean                  adaptivePtime = false;
+    RTCEncodingComplexity    encodingComplexity;
 };</pre>
     <section id="rtcrtpencodingparameters-attributes">
       <h2>Dictionary {{RTCRtpEncodingParameters}} Members</h2>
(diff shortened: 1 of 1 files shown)


youennf:
Given this might be even more useful to WebCodecs encoders, it seems we should discuss this in WebCodecs repo and maybe piggy back on what is added there in terms of enum for instance.

youennf:
@sprangerik, do you plan to file an issue to WebCodecs?

sprangerik:
> @sprangerik, do you plan to file an issue to WebCodecs?

I do plan to file an issue with WebCodecs - but I think that implementation will look quite different. This enum is a high-level hint to UA and then it does its best. 

WebCodecs is a low-level API in comparison, and there this type of hint doesn't make much sense. Instead, we'll need a way to figure out what the capabilities of the encoder are (i.e. the range of valid speed settings) so that a user can know that changing the value will actually change the behavior. I also think that parameter should be done on a per-frame bases in WebCodecs, which makes no sense for WebRTC.

jan-ivar:
I'd expect a higher-level API to need less granularity not more.

sprangerik:
WebRTC (the higher-level) API would indeed have less granularity - as in these ~5 or so enum "hint" values.

(I've been swamped by other things, but will upload a change to reduce the number of values soon, as discussed.)

WebCodecs on the other hand should have a strict mapping to underlying implementation - so e.g. libvpx VP8 would expose the full 1..12 range. OpenH265 would have 1..3. That means we also need a way query what settings are available for the encoder instance in question.

PR #255, opened 2026-07-28
Add section for Diagnostic Logging API

guidou:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-extensions/pull/255.html" title="Last updated on Sep 11, 2026, 11:58 AM UTC (dd50f2d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-extensions/255/fb2c4cd...dd50f2d.html" title="Last updated on Sep 11, 2026, 11:58 AM UTC (dd50f2d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1808,7 +1808,7 @@ partial interface RTCRtpTransceiver {
       applications to start, finish, and cancel the collection of internal diagnostic
       logs for WebRTC-related operations performed by the user agent.
       These diagnostic logs are never exposed to the application. Instead they are
-      stored in a location controlled by the user and maybe shared with the
+      stored in a location controlled by the user and may be shared with the
       user agent vendor.
       The collection and storage of diagnostic logs requires explicit
       authorization by user and the application never knows if diagnostic logging
--- a/index.html
+++ b/index.html
@@ -1802,6 +1802,310 @@ partial interface RTCRtpTransceiver {
       </table>
     </section>
   </section>
+  <section id="diagnostic-logging">
+    <h2>Diagnostic logging</h2>
+      The WebRTC Diagnostic Logging API provides a programmatic interface for web
+      applications to start, finish, and cancel the collection of internal diagnostic
+      logs for WebRTC-related operations performed by the user agent.
+      These diagnostic logs are never exposed to the application. Instead they are
+      stored locally by the user agent and are under the control of the user.
+      The user agent can also upload the diagnostic logs to an endpoint decided by
+      the user agent.
+      The collection, storage and upload of diagnostic logs requires explicit
+      authorization by user and the application never knows if these operations
+      succeed. The contents of the diagnostic logs are also an implementation
+      detail.
+      The use cases intended to be supported by this API are the following:
+      <ul>
+        <li>
+          An application requests the collection of logs, which are stored locally. 
+          These logs can be used by a developer to help diagnose bugs in the
+          application. A user of the application can provide these logs to an
+          application developer in order to help fix bugs or otherwise improve the
+          application. An organization can collect these logs from its users to
+          diagnose bugs or make improvements.
+        </li>
+        <li>
+          An application may request that the logs be shared with the user agent vendor.
+          This is useful in case the application developer suspects a bug in the user
+          agent and it wants to provide the logs to help the user agent developer fix
+          the bug. To support this use case, the API returns a UUID that can be included
+          in a bug report to identify an uploaded diagnostic log. This mechanism allows
+          a user to authorize exposing the diagnostic log to the user agent vendor,
+          but it does not allow exposing the diagnostic log to the application.
+        </li>
+      </ul>
+      <section id="diagnostic-logging-security-privacy">
+        <h3>Security and Privacy</h3>
+        These diagnostic logs collect information about internal operations performed
+        by the user agent to implement WebRTC-related features. These diagnostic logs
+        may contain information not exposed to the Web application and therefore, the
+        API cannot expose these logs to the Web application in any way. The logs may,
+        subject to user authorization, be shared (via out-of-band uploads, for example)
+        with the user-agent vendor so that they can be used to assist in fixing
+        user-agent bugs or providing other improvements to the user agent.
+
+        The collection, storage and upload of diagnostic logs requires explicit
+        authorization by the user. The specific mechanism for this authorization is
+        [=implementation-defined=]. Some options to implement authorization include, but are
+        not limited to, dedicated UI, settings, enterprise policies, prompts, or
+        combination thereof. Authorization may be limited to specific origins. The
+        status of these authorizations is never exposed to the application. Therefore,
+        the APIs provide no guarantees of success.
+      </section>
+      <section id="diagnostic-logging-extensions-peer-connection">
+        <h3>Extensions to the RTCPeerConnection Interface</h3>
+        The API is exposed as a set of static methods on the {{RTCPeerConnection}}
+        interface.
+        <pre class="idl">
+          [
+            Exposed=Window
+          ] partial interface RTCPeerConnection {
+            static Promise&lt;DOMString&gt; startDiagnosticLogging(optional RTCStartDiagnosticLoggingOptions options = {});
+            static Promise&lt;undefined&gt; finishDiagnosticLogging(optional RTCFinishDiagnosticLoggingOptions options = {});
+            static Promise&lt;undefined&gt; cancelDiagnosticLogging();
+          };
+
+          dictionary RTCDiagnosticLoggingOptions {
+            record&lt;DOMString, DOMString&gt; metadata;
+          };
+
+          dictionary RTCStartDiagnosticLoggingOptions : RTCDiagnosticLoggingOptions {
+            boolean allowUpload = false;
+          };
+
+          dictionary RTCFinishDiagnosticLoggingOptions : RTCDiagnosticLoggingOptions {
+          };
+        </pre>
+      </section>
+      <section id="diagnostic-logging-internal-slots">
+        <h3>Internal slots</h3>
+          Let the the [=relevant global object=] have an
+          <dfn data-dfn-for="RTCPeerConnection">[[\RTCDiagnosticLoggingSessionId]]</dfn>
+          internal slot, initialized to <code>null</code>.
+      </section>
+      <section id="diagnostic-logging-methods">
+        <h3>Methods</h3>
+        <dl data-link-for="RTCPeerConnection" data-dfn-for="RTCPeerConnection" class="methods">
+          <dt>
+            <dfn>startDiagnosticLogging</dfn>
+          </dt>
+          <dd>
+            <p>The {{startDiagnosticLogging}} method attempts to start a diagnostic
+               logging session.</p>
+            <p>When this method is invoked, the [= user agent =] MUST run the following steps:</p>
+            <ol class="algorithm">
+              <li>
+                <p>Let |options| be the first argument of this method.</p>
+              </li>
+              <li>
+                <p>Let |allowUpload| be |options|'s
+                  {{RTCStartDiagnosticLoggingOptions/allowUpload}} member.</p>
+              </li>
+              <li>
+                <p>Let |metadata| be |options|'s
+                  {{RTCDiagnosticLoggingOptions/metadata}} member.</p>
+              </li>
+              <li>
+                <p>If the size of |metadata| exceeds 5 entries, or if any key or
+                   value in |metadata| exceeds 100 characters, return a promise
+                   [=rejected=] with a {{TypeError}}.</p>
+              </li>
+              <li>
+                <p>Let |p| be a new promise.</p>
+              </li>
+              <li>
+                <p>[=In parallel=], perform the following steps:</p>
+                <ol>
+                  <li>
+                    <p>Let |uuid| be a randomly generated UUID [[rfc4122]], which
+                       is 36 characters long in its canonical form. To avoid
+                       fingerprinting, implementations SHOULD use the forms in
+                       section 4.4 or 4.5 of RFC 4122 when generating UUIDs.</p>
+                  </li>
+                  <li>
+                    <p>Let |doc| be the [=relevant global object=]'s associated [=document=].</p>
+                  </li>
+                  <li>
+                    <p>If |doc|'s [=browsing context=] is not a [=top-level browsing context=],
+                       [=resolve=] |p| with |uuid| and abort these steps.</p>
+                  </li>
+                  <li>
+                    <p>If the {{RTCPeerConnection/[[RTCDiagnosticLoggingSessionId]]}} internal slot
+                       is not <code>null</code>, [=resolve=] |p| with |uuid| and abort these steps.</p>
+                  </li>
+                  <li>
+                    <p>Store |uuid| in the {{RTCPeerConnection/[[RTCDiagnosticLoggingSessionId]]}}
+                       internal slot.</p>
+                  </li>
+                  <li>
+                    <p>[=Resolve=] |p| with |uuid|.</p>
+                  </li>
+                  <li>
+                    <p>Start a logging session of internal WebRTC activity identified with |uuid|</p>
+                  </li>
+                </ol>
+              </li>
+              <li>
+                <p>Return |p|.</p>
+              </li>
+            </ol>
+            <p>
+              Once a logging session starts, any WebRTC-related activity generated
+              by |doc| or its descendant documents that are of the [=same origin=]
+              as |doc| MAY be logged by the user agent after |p| resolves.
+              The log MAY include |metadata| or information derived from it.
+              If |allowUpload| is <code>true</code>, the logged data may be shared with the
+              user agent vendor via an [=implementation-defined=] mechanism, as long as the user
+              has authorized it and the logging session has not been cancelled with the
+              {{cancelDiagnosticLogging}} method. The logging session is identified with
+              |uuid|, which means that all logged data can be internally referenced using
+              |uuid|.
+            </p>
+          </dd>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1802,6 +1802,310 @@ <h3>
...
   </section>
+  <section id="diagnostic-logging">
+    <h2>Diagnostic logging</h2>
+      The WebRTC Diagnostic Logging API provides a programmatic interface for web
+      applications to start, finish, and cancel the collection of internal diagnostic
+      logs for WebRTC-related operations performed by the user agent.
+      These diagnostic logs are never exposed to the application. Instead they are
+      stored locally by the user agent and are under the control of the user.
+      The user agent can also upload the diagnostic logs to an endpoint decided by
+      the user agent.
```
jan-ivar, line 1813: This seems too broad. Can we scope it more clearly?

index.html
```diff
@@ -1802,6 +1802,310 @@ <h3>
...
+      The collection, storage and upload of diagnostic logs requires explicit
+      authorization by user and the application never knows if these operations
```
jan-ivar, line 1815: Can we clarify how the user agent is supposed to obtain explicit authorization? Would this involve UX?

If we have such UX do we then also need `allowUpload` boolean?

index.html
guidou: Updated the wording to make it clear that they are stored in a location specified by the user.

index.html
guidou: It is explained elsewhere that it can be browser UX, enterprise policies, etc.

I agree that `allowUpload` is unnecessary because the user controls where the logs are stored, so `allowUpload` in practice is always true from an API point of view. The user decides using browser UI/config/policies where the logs go.

index.html
```diff
@@ -1802,6 +1802,267 @@ <h3>
...
+                <p>Let |p| be a new promise.</p>
+              </li>
+              <li>
+                <p>[=In parallel=], perform the following steps:</p>
+                <ol>
+                  <li>
+                    <p>Let |uuid| be a randomly generated UUID [[rfc4122]], which
+                       is 36 characters long in its canonical form. To avoid
+                       fingerprinting, implementations SHOULD use the forms in
+                       section 4.4 or 4.5 of RFC 4122 when generating UUIDs.</p>
```
youennf: I would tend to refer to section 4.4 only, which is version 4 UUID.

index.html
```diff
@@ -1802,6 +1802,267 @@ <h3>
...
+                  </li>
+                </ol>
+              </li>
+              <li>
+                <p>Return |p|.</p>
+              </li>
+            </ol>
+            <p>
+              Once a logging session starts, any WebRTC-related activity generated
+              by |doc| or its descendant documents that are of the [=same origin=]
```
youennf: I would simplify to |doc| only and leave descendant handling to JS.

index.html
```diff
@@ -1802,6 +1802,267 @@ <h3>
...
+              <li>
+                <p>Let |options| be the first argument of this method.</p>
+              </li>
+              <li>
+                <p>Let |metadata| be |options|'s
+                  {{RTCDiagnosticLoggingOptions/metadata}} member.</p>
+              </li>
+              <li>
+                <p>If the size of |metadata| exceeds 5 entries, or if any key or
+                   value in |metadata| exceeds 100 characters, return a promise
```
jan-ivar: "characters" is ambiguous. An implementation checking str.length and one checking [...str].length will disagree on whether the same call throws. That's a WPT-visible interop bug.

Fix: use Infra's terms, which are defined for exactly this.

If |metadata|'s [=map/size=] is greater than 5, or if any [=map/entry=]
|key| → |value| of |metadata| has a [=string/length=] greater than 100,
return a promise [=rejected=] with a {{TypeError}}.

[=string/length=] is code units — it matches str.length, so it's the cheapest to implement and the easiest to test. If you'd rather bound storage, say UTF-8 bytes instead, since 100 code units can be 400 bytes. But pick one and name it.

index.html
```diff
@@ -1802,6 +1802,267 @@ <h3>
...
+                  </li>
+                  <li>
+                    <p>Let |doc| be the [=relevant global object=]'s associated [=document=].</p>
+                  </li>
+                  <li>
+                    <p>If |doc|'s [=browsing context=] is not a [=top-level browsing context=],
+                       [=resolve=] |p| with |uuid| and abort these steps.</p>
+                  </li>
+                  <li>
+                    <p>If the {{RTCPeerConnection/[[RTCDiagnosticLoggingSessionId]]}} internal slot
```
youennf: If start is called twice, I would tend to have the second reject.

index.html
```diff
@@ -1802,6 +1802,267 @@ <h3>
...
+                    <p>[=Resolve=] |p| with <code>undefined</code>.</p>
+                  </li>
+                </ol>
+              </li>
+              <li>
+                <p>Return |p|.</p>
+              </li>
+            </ol>
+            <p>
+              The user agent MUST not log any WebRTC-related activity generated by |doc| or
```
jan-ivar:
```suggestion
              The user agent MUST NOT log any WebRTC-related activity generated by |doc| or
```

index.html
```diff
@@ -1802,6 +1802,267 @@ <h3>
...
+              its descendant documents after the |p| resolves. The log MAY
+              include |metadata| or information derived from it.
```
jan-ivar: Said elsewhere
```suggestion
              its descendant documents after the |p| resolves.
```

index.html
```diff
@@ -1802,6 +1802,267 @@ <h3>
...
+              </li>
+              <li>
+                <p>Return |p|.</p>
+              </li>
+            </ol>
+            <p>
+              After |p| resolves:
+              <ul>
+                <li>
+                  The user agent MUST not log any WebRTC-related activity
```
jan-ivar:
```suggestion
                  The user agent MUST NOT log any WebRTC-related activity
```

index.html
```diff
@@ -1802,6 +1802,267 @@ <h3>
...
+              The user agent MUST not log any WebRTC-related activity generated by |doc| or
+              its descendant documents after the |p| resolves. The log MAY
+              include |metadata| or information derived from it.
+            </p>
+          </dd>
+          <dt>
+            <dfn>cancelDiagnosticLogging</dfn>
+          </dt>
+          <dd>
+            <p>The {{cancelDiagnosticLogging}} method cancels an ongoing diagnostic logging session.</p>
```
jan-ivar: Maybe
```suggestion
            <p>The {{cancelDiagnosticLogging}} method cancels an ongoing diagnostic logging session and deletes the log.</p>
```

jan-ivar (approved):
Overall seems good with nits and @youennf's comment on rejecting if uuid internal slot is not null.

index.html
guidou: Applied the suggestion with minor wording change and incorporating the updated method name.

index.html
guidou: Done

index.html
guidou: Done

index.html
guidou: Done

index.html
guidou: Done

index.html
guidou: Done

index.html
guidou: Done

index.html
guidou: done

index.html
```diff
@@ -1819,6 +1802,277 @@ <h3>
...
+                       a {{DOMException}} object whose {{DOMException/name}}
+                       attribute has the value {{"InvalidStateError"}} and abort
+                       these steps.</p>
+                  </li>
+                  <li>
+                    <p>Store |uuid| in the {{RTCPeerConnection/[[RTCDiagnosticLoggingSessionId]]}}
+                       internal slot.</p>
+                  </li>
+                  <li>
+                    <p>[=Resolve=] |p| with |uuid|.</p>
```
youennf: We usually queue a task to resolve when doing in parallel.

index.html
```diff
@@ -1819,6 +1802,277 @@ <h3>
...
+              <li>
+                <p>Let |p| be a new promise.</p>
+              </li>
+              <li>
+                <p>[=In parallel=], perform the following steps:</p>
+                <ol>
+                  <li>
+                    <p>Let |uuid| be a randomly generated UUID [[rfc4122]]. To avoid
+                       fingerprinting, implementations SHOULD use the form in
+                       section 4.4 of RFC 4122 when generating UUIDs.</p>
```
youennf: Would best be done synchronously, similar to the InvalidStateError error.

index.html
```diff
@@ -1819,6 +1802,277 @@ <h3>
...
+      </section>
+      <section id="diagnostic-logging-extensions-peer-connection">
+        <h3>Extensions to the RTCPeerConnection Interface</h3>
+        The API is exposed as a set of static methods on the {{RTCPeerConnection}}
+        interface.
+        <pre class="idl">
+          [
+            Exposed=Window
+          ] partial interface RTCPeerConnection {
+            static Promise&lt;DOMString&gt; startDiagnosticLogging(optional RTCDiagnosticLoggingOptions options = {});
```
youennf: Would be good without a promise. It removes a potential timing issue if starting logging takes time and JS can identify when logging happens and when not.

index.html
```diff
@@ -1819,6 +1802,277 @@ <h3>
...
+            static Promise&lt;undefined&gt; finishDiagnosticLogging(optional RTCDiagnosticLoggingOptions options = {});
+            static Promise&lt;undefined&gt; discardDiagnosticLogging();
```
jan-ivar: "Discard" sounds like it can be called past finish. I'd prefer:
```suggestion
            static Promise&lt;undefined&gt; cancelDiagnosticLogging();
```

index.html
```diff
@@ -1819,6 +1802,277 @@ <h3>
...
+            <p>The {{startDiagnosticLogging}} method attempts to start a diagnostic
+               logging session.</p>
+            <p>When this method is invoked, the [= user agent =] MUST run the following steps:</p>
+            <ol class="algorithm">
+              <li>
+                <p>Let |options| be the first argument of this method.</p>
+              </li>
+              <li>
+                <p>Let |metadata| be |options|'s
+                  {{RTCDiagnosticLoggingOptions/metadata}} member.</p>
```
jan-ivar: A dictionary [member](https://webidl.spec.whatwg.org/#dfn-dictionary-member) is technically the key value tuple. To get the value just write:
```suggestion
                  {{RTCDiagnosticLoggingOptions/metadata}}.</p>
```

index.html
```diff
@@ -1819,6 +1802,277 @@ <h3>
...
+              </li>
+              <li>
+                <p>If |metadata|'s [=map/size=] is greater than 5, or if any
+                   [=map/entry=] |key| → |value| of |metadata| has a 
+                   [=string/length=] greater than 100, return a promise
```
jan-ivar: Does this mean key < 100 && value < 100 or (key + value) < 100? If the former, how about:
```suggestion
                <p>If |metadata|'s [=map/size=] is greater than 5, or
                   |metadata| [=map/contains=] any [=map/entry=] whose
                   [=map/key=] or [=map/value=] have a [=string/length=]
                   greater than 100, return a promise
```

index.html
```diff
@@ -1819,6 +1802,277 @@ <h3>
...
+                  </li>
+                  <li>
+                    <p>Store |uuid| in the {{RTCPeerConnection/[[RTCDiagnosticLoggingSessionId]]}}
+                       internal slot.</p>
+                  </li>
+                  <li>
+                    <p>[=Resolve=] |p| with |uuid|.</p>
+                  </li>
+                  <li>
+                    <p>Start a logging session of internal WebRTC activity identified with |uuid|</p>
```
jan-ivar: I'd make everything else synchronous (no start promise) and do just this one line in parallel.

Also, this algorithm doesn't do anything with _metadata_ so let's reference it here.

```suggestion
                    <p>[=In parallel=], start a logging session of internal WebRTC
                    activity associated with |uuid|, using |metadata| for context.</p>
```

index.html
```diff
@@ -1819,6 +1802,277 @@ <h3>
...
+                <p>Let |metadata| be |options|'s {{RTCDiagnosticLoggingOptions/metadata}} member.</p>
+              </li>
+              <li>
+                <p>If |metadata|'s [=map/size=] is greater than 5, or if any
+                   [=map/entry=] |key| → |value| of |metadata| has a 
+                   [=string/length=] greater than 100, return a promise
+                   [=rejected=] with a {{TypeError}}.
+                  </p>
+              </li>
+              <li>
```
jan-ivar: metadata is unused here. Remove options

index.html
```diff
@@ -1819,6 +1802,277 @@ <h3>
...
+      <section id="diagnostic-logging-extensions-peer-connection">
+        <h3>Extensions to the RTCPeerConnection Interface</h3>
+        The API is exposed as a set of static methods on the {{RTCPeerConnection}}
+        interface.
+        <pre class="idl">
+          [
+            Exposed=Window
+          ] partial interface RTCPeerConnection {
+            static Promise&lt;DOMString&gt; startDiagnosticLogging(optional RTCDiagnosticLoggingOptions options = {});
+            static Promise&lt;undefined&gt; finishDiagnosticLogging(optional RTCDiagnosticLoggingOptions options = {});
```
jan-ivar: I find not precedence for start()/finish() in the web platform.

The proposed algorithm itself says “Stop the logging session”. It doesn’t finalize anything. So I’d suggest:

```suggestion
            static Promise&lt;undefined&gt; stopDiagnosticLogging();
```
The options argument appears unused so we should drop it (there can only be one session running at a time).

index.html
```diff
@@ -1819,6 +1802,277 @@ <h3>
...
+                       these steps.</p>
+                  </li>
+                  <li>
+                    <p>Stop the logging session identified with 
+                      {{RTCPeerConnection/[[RTCDiagnosticLoggingSessionId]]}}.</p>
+                  </li>
+                  <li>
+                    <p>Set {{RTCPeerConnection/[[RTCDiagnosticLoggingSessionId]]}}
+                       to <code>null</code>.</p>
+                  </li>
```
jan-ivar: I'd move the check and reset of the internal slot up to the synchronous part, so all manipulation of the internal slot is on the main thread, and leave only the "Stop the logging session" in parallel, plus queue a task to resolve the promise.

index.html
```diff
@@ -1819,6 +1802,277 @@ <h3>
...
+                       a {{DOMException}} object whose {{DOMException/name}}
+                       attribute has the value {{"InvalidStateError"}} and discard
+                       these steps.</p>
+                  </li>
+                  <li>
+                    <p>Let |uuid| be the value of the {{RTCPeerConnection/[[RTCDiagnosticLoggingSessionId]]}}.</p>
+                  </li>
+                  <li>
+                    <p>Discard the logging session identified with 
+                      {{RTCPeerConnection/[[RTCDiagnosticLoggingSessionId]]}}.</p>
```
jan-ivar: Same here.

Also _uuid_ is unused.

jan-ivar (changes requested):

index.html
guidou: Done.

index.html
guidou: Done

index.html
guidou: Done

index.html
guidou: Done

index.html
guidou: Done

index.html
guidou: Done

index.html
guidou: Done

index.html
guidou: Done.

index.html
guidou: Done

index.html
guidou: Done

index.html
```diff
@@ -1819,6 +1802,274 @@ <h3>
...
+              <li>
+                <p>If |metadata|'s [=map/size=] is greater than 5, or if any
+                   [=map/entry=] |key| → |value| of |metadata| has a 
+                   [=string/length=] greater than 100, return a promise
+                   [=rejected=] with a {{TypeError}}.
+                  </p>
+              </li>
+              <li>
+                <p>If the {{RTCPeerConnection/[[RTCDiagnosticLoggingSessionId]]}}
+                    internal slot is not <code>null</code>, [=reject=] |p| with
```
youennf: We no longer reject, we should throw now.

index.html
```diff
@@ -1819,6 +1802,274 @@ <h3>
...
+              </li>
+              <li>
+                <p>Let |metadata| be |options|'s
+                  {{RTCDiagnosticLoggingOptions/metadata}}.</p>
+              </li>
+              <li>
+                <p>If |metadata|'s [=map/size=] is greater than 5, or if any
+                   [=map/entry=] |key| → |value| of |metadata| has a 
+                   [=string/length=] greater than 100, return a promise
+                   [=rejected=] with a {{TypeError}}.
```
youennf: Ditto here, we should throw a TypeError.

jan-ivar (approved):
Approve with throw

youennf (approved):
LGTM.
I would with no promise for stop/cancel

henbos:
Make sync but editors can integrate

jan-ivar:
Editors can integrate once synchronous

index.html
guidou: Done

index.html
guidou: Done

index.html
guidou: Done

PR #256, opened 2026-07-29
rtp header encryption changes

fippo:
as proposed in
  https://www.w3.org/2026/06/16-webrtc-minutes.html#9a7b

(finally got around to doing it)

***
<a href="https://pr-preview.s3.amazonaws.com/fippo/webrtc-extensions/pull/256.html" title="Last updated on Sep 9, 2026, 2:12 PM UTC (29f92c6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-extensions/256/fb2c4cd...fippo:29f92c6.html" title="Last updated on Sep 9, 2026, 2:12 PM UTC (29f92c6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1363,8 +1363,8 @@ dictionary RTCEncodingOptions {
               <td>
                 <p>
                   Do not use RTP header encryption. The "a=cryptex" attribute is not
-                  included in offers or answers, and RTP header extensions and CSRCs are
-                  sent in the clear.
+                  included in offers or answers and not acted upon when present in the
+                  remote description. RTP header extensions and CSRCs are sent in the clear.
                 </p>
               </td>
             </tr>
@@ -1375,7 +1375,9 @@ dictionary RTCEncodingOptions {
                   Negotiate RTP header encryption as defined in [[RFC9335]]. If the remote
                   endpoint does not indicate that it is capable of receiving RTP packets
                   encrypted with cryptex, RTP header extensions and CSRCs are sent in the
-                  clear.
+                  clear. If the remote endpoint indicates support for cryptex, the user
+                  agent may use cryptex for endcrypting RTP header extensions and CSRCS
+                  as described in [[RFC9335]] Section 4.
                 </p>
               </td>
             </tr>
--- a/index.html
+++ b/index.html
@@ -1435,7 +1435,7 @@ dictionary RTCEncodingOptions {
             </p>
             <p class="needs-test">
               {{RTCConfiguration/rtpHeaderEncryptionPolicy}} can only be set when
-              constructing the {{RTCPeerConnection}} and MUST NOT be changed afterwards.
+              constructing the {{RTCPeerConnection}} and cannot be changed afterwards.
               In the <a data-cite="WEBRTC#set-the-configuration">set the configuration</a>
               algorithm, add a step after the step saying
               "If the value of configuration.rtcpMuxPolicy differs from oldConfig.rtcpMuxPolicy, then fail."
--- a/index.html
+++ b/index.html
@@ -1327,19 +1327,28 @@ dictionary RTCEncodingOptions {
   </section>
 </section>
   <section id="rtp-header-extension-encryption">
-    <h3>RTP Header Extension Encryption</h3>
+    <h3>RTP Header Encryption</h3>
     <section id="rtp-header-extension-encryption-policy">
       <h3>
           <dfn>RTCRtpHeaderEncryptionPolicy</dfn> Enum
       </h3>
       <p>
-        RTP header extension encryption policy affects whether RTP header extension
-        encryption is negotiated if the remote endpoint does not support [[RFC9335]].
-        If the remote endpoint supports [[RFC9335]], all media streams are sent
-        utilizing [[RFC9335]].
+        [[RFC9335]], also known as cryptex, defines a mechanism for encrypting the RTP
+        header extensions and the contributing source (CSRC) identifiers of an RTP packet,
+        both of which are left in the clear by [[RFC3711]]. It is signaled with the
+        "a=cryptex" SDP attribute, whose presence in an offer or an answer indicates that
+        the endpoint is capable of receiving RTP packets encrypted with cryptex. Once both
+        endpoints have indicated that capability, a sender decides on a per-packet basis
+        whether to apply cryptex.
+      </p>
+      <p>
+        The RTP header encryption policy determines whether the user agent indicates that
+        capability, and whether it is willing to continue with a remote endpoint which has
+        not indicated it.
       </p>
       <div>
         <pre id="target-rtp-header-encryption-policy" class="idl">enum RTCRtpHeaderEncryptionPolicy {
+  "disable",
   "negotiate",
   "require"
 };</pre>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1350,68 +1359,50 @@ <h3>
...
+                  {{RTCConfiguration/rtpHeaderEncryptionPolicy}} is
+                  {{RTCRtpHeaderEncryptionPolicy/require}} and the description does not
+                  indicate that the remote endpoint is capable of receiving RTP packets
+                  encrypted with cryptex, then [= reject =] <var>p</var>
                   with a newly [= exception/created =] {{InvalidAccessError}} and abort these steps.
                 </p>
+                <p>
+                  In addition, the user agent MUST consider the use of [[RFC9335]]
+                  mandatory, as described in [[RFC9335]] Section 5.2. Received RTP packets
+                  which are not encrypted with cryptex are not processed further.
```
youennf, line 1398: Is it clear how these dropped packets would appear on webrtc-stats? Will they even appear?

index.html
```diff
@@ -1350,68 +1359,50 @@ <h3>
...
+              </td>
+            </tr>
+            <tr>
+              <td><dfn data-idl>negotiate</dfn></td>
+              <td>
                 <p>
+                  Negotiate RTP header encryption as defined in [[RFC9335]]. If the remote
+                  endpoint does not indicate that it is capable of receiving RTP packets
+                  encrypted with cryptex, RTP header extensions and CSRCs are sent in the
+                  clear.
```
youennf, line 1378: What happens in case the remote endpoint supports cryptex in that case?
Now that we have `disable`, I would think that the UA MUST send cryptex packets.
If it cannot be a MUST, a SHOULD would be good.

youennf:
Overall looks good to me.

index.html
fippo: we don't have a SRTP unprotect failure counter so they would be dropped silently. Might be a useful thing to have regardless (I am looking at you error=9)

index.html
fippo: Good point. The current definition of `disable` leans towards "don't use it" but that needs text clarifying that it is supposed to ignore cryptex in the remote description too.
I'm generally in favor of "opportunistic cryptography" so :+1 for adding a "if the remote offers, you may use it" 

If `disable` stays such a opt-out it might be good to mark it `at risk` upfront to make it clear we aim for more crypto.

(the case where the offerer is configured with `disable` and the answerer with `require` would break in both cases I think, it just depends how badly... don't shoot yourself in the foot)

jan-ivar: So something like this?

```suggestion
                  clear.
                  If the remote endpoint indicates that it is capable of
                  receiving RTP packets encrypted with cryptex, the
                  user agent SHOULD apply cryptex. 
```

index.html
```diff
@@ -1436,6 +1427,28 @@ <h2>
...
             <p class="needs-test">
+              The {{RTCConfiguration/rtpHeaderEncryptionPolicy}} applies to the
+              {{RTCPeerConnection}} as a whole. The "a=cryptex" attribute is assigned to
+              the TRANSPORT category [[RFC8859]], so when BUNDLE is in use it is either
+              present on every RTP-based m= section of a bundle group or on none of them,
+              as required by [[RFC9335]] Section 4.
+            </p>
+            <p class="needs-test">
+              {{RTCConfiguration/rtpHeaderEncryptionPolicy}} can only be set when
+              constructing the {{RTCPeerConnection}} and MUST NOT be changed afterwards.
```
jan-ivar: The set algorithm below is clear so we can say
```suggestion
              constructing the {{RTCPeerConnection}} and cannot be changed afterwards.
```

index.html
```diff
@@ -1436,6 +1427,28 @@ <h2>
...
+              saying
+              "If the value of configuration.{{RTCConfiguration/rtpHeaderEncryptionPolicy}} differs from
+              oldConfig.{{RTCConfiguration/rtpHeaderEncryptionPolicy}}, then fail."
+            </p>
+            <p>
+              In setConfiguration in [[RTCWEB-JSEP]] section 4.1.18, replace
+              "The bundle and RTCP-multiplexing policies MUST NOT be changed after the construction of the PeerConnection."
+              with
+              "The bundle policy, the RTCP-multiplexing policy and the RTP header encryption policy
+              MUST NOT be changed after the construction of the PeerConnection."
```
jan-ivar, line 1451: We probably don't need to modify JSEP to add an additional restriction. Might it suffice to say:

```suggestion
              This spec adds the following restriction above and beyond what is described for
               setConfiguration in [[RTCWEB-JSEP]] section 4.1.18:
              "The RTP header encryption policy MUST NOT be changed after the construction of the PeerConnection."
```
Either way.

jan-ivar (approved):

index.html
fippo: I do not think we should narrow cryptex [which says](https://www.rfc-editor.org/rfc/rfc9335.html#section-4-4)
> Once each peer has verified that the other party supports receiving RTP packets encrypted with Cryptex, senders can unilaterally decide whether or not to use the Cryptex mechanism on a per-packet basis.

index.html
fippo: We always did however and also do in 15.3 - https://w3c.github.io/webrtc-extensions/#always-negotiating-datachannels-interface-extensions

fippo:
ping, anything needed from me?

index.html
youennf: This spec is trying to make use of cryptex in web browsers, narrowing it down is ok to me, at least on the sender side. Otherwise, we can easily get interop issues.

Do we have a usecase for web browsers to not use cryptex once negotiated?
@fippo, is libwebrtc implementation sometimes opting out of sending with cryptex when negotiated?

fippo: >  we can easily get interop issues

Can you give an example?

> is libwebrtc implementation sometimes opting out of sending with cryptex when negotiated?

No but why narrow the specification without need and restricting IETF consensus?

index.html
jan-ivar: A formal change to JSEP must go through the IETF. cc @docfaraday

15.3 moves the data m= section from "lastly" to first, which genuinely contradicts RFC 9429 §5.2.1. It was added in https://github.com/w3c/webrtc-extensions/pull/242 which at least opened https://github.com/rtcweb-wg/jsep/issues/1039 asking for this in _"the next revision"_ and remains open. If we need non-browser endpoints to change we need to do the same here.

In contrast, this PR adds an additional restriction above and beyond the restrictions in JSEP, which seems fine.

A distinction to make:
1. Browser endpoints (User Agents) MUST follow requirements in _both_ JSEP _AND_ W3C specs
2. Non-browser endpoints MUST follow JSEP, but W3C specs do not apply

[createOffer](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-createoffer) for example says:
- _"Creating the SDP MUST follow … RFC 9429, except the user agent MUST [additional requirement]..."_

index.html
jan-ivar: The need would be to ensure user privacy and interop by having browsers behave the same.

Are there any downsides? A SHOULD shouldn't get in the way of incremental adoption. It's easier to relax requirements later than restrict them later, so starting narrow makes sense to me.

What IETF consensus is violated? The quoted text says _"senders can unilaterally decide whether or not to use the Cryptex mechanism on a per-packet basis."_

For browser endpoints, the _"sender"_ is the user agent.

This would seem to suggest we can decide this unilaterally for browser endpoints.

index.html
fippo: filed https://github.com/rtcweb-wg/jsep/issues/1041 since this is relevant for native applications too.

docfaraday: I am of the opinion that modifying JSEP for this is unnecessary; putting a "Do not change" restriction in w3c is sufficient. At the risk of soapboxing, JSEP should never have mentioned the setConfiguration JS API surface in the first place, and should have delegated these choices to the app layer like every other IETF spec.

fippo: the split was always weird. That one implementation sits inside libWebRTC's C++ rather than Chromium's glue suggests into is below the W3C layer though.

index.html
youennf: > Can you give an example?

Whenever there is a difference of behavior, there is an interop issue, we saw it so many times in WebRTC that I prefer we avoid it here.
If there is a good reason to not always use cryptex, describing the conditions where this happens is healthy for the community and will help future implementors.
The place to describe this could be this spec if it is UA-specific or IETF it is generic to all cryptex impls.

> why narrow the specification without need and restricting IETF consensus?

It is more about describing how a technology is used in one specific context (UAs) than narrowing.

fippo: > Are there any downsides?

A "SHOULD" does not close a loophole. `negotiate` already allows renegotiation to disable cryptex. This is not per-packet but per-negotiation but it is the same result with less granularity. If you want to avoid that use `require`.

Anyway, text updated.

youennf (approved):

jan-ivar (approved):
LLGTM. The JSEP part seems harmless to me

